PR #29, opened 2026-04-26
feat(bottom_bar): opt-in indicator behind icons via indicatorBehindIcons

jfhair:
## Summary

Adds a non-breaking `indicatorBehindIcons: bool = false` flag to `GlassBottomBar` and `GlassSearchableBottomBar` that swaps the Z-order of the moving glass indicator and the icon content layer.

- **Default (`false`)**: unchanged — indicator paints *on top* of icons (the current "magic lens" refraction behaviour).
- **Opt-in (`true`)**: indicator paints *behind* icons. Matches iOS-native tab bar (`UITabBar`) where the active-state pill is a background layer and icons render on top.

## Motivation

When using a saturated `selectedIconColor` (e.g. iOS `systemBlue`), the indicator's translucent white glass lens visibly washes the active icon's colour out. The icon ends up looking grey/muted in its selected state, even with `indicatorColor` set very low. This is because the indicator is currently the *topmost* layer in the Stack (lines 528–551 in `bottom_bar_internal.dart` and the equivalent spots in the high-quality builder + searchable variant) — anything translucent painted on top of an icon dims its colour.

iOS's own `UITabBar` paints the indicator pill behind the icon, which is why an Apple `systemBlue` icon stays vivid even when selected. Apps that want this behaviour currently have no way to get it without forking.

## Approach

- Added a `final bool indicatorBehindIcons` field (default `false`) to both public widgets and both internal `*TabIndicator` widgets.
- Plumbed the value through public → internal.
- In each of the four Stack builders (simple + high-quality × bottom-bar + searchable-bottom-bar), the icons-layer and indicator-layer are now extracted into local `final` variables and conditionally ordered:

```dart
if (widget.indicatorBehindIcons) ...[
  indicatorLayer,
  iconsLayer,
] else ...[
  iconsLayer,
  indicatorLayer,
],
```

No public layout/sizing shifts, no new dependencies. The lens-refraction effect is preserved as the default; only opted-in apps see the new ordering.

## Test plan

- [x] `flutter analyze` clean (no new errors / warnings)
- [x] Tested with `dependency_overrides` in a real app (Hero Dirt — bottom-bar with searchable variant + 3 tabs + saturated active-tab tint). Default `false` matches main; `true` resolves the icon-washing issue cleanly.
- [ ] Maintainer to spot-check that animated transitions (jelly clip, spring-driven indicator move) still look correct in both modes — they should, since the indicator and icon widgets are unchanged; only their position in the parent Stack differs.

## Notes

- I'm happy to add a CHANGELOG entry / docs example / test if useful — let me know what you'd like.
- If a different name (e.g. `indicatorBelowIcons`, `iconsAboveIndicator`) reads better, easy to rename.

### DIFF
--- a/lib/widgets/surfaces/glass_bottom_bar.dart
+++ b/lib/widgets/surfaces/glass_bottom_bar.dart
@@ -195,6 +195,7 @@ class GlassBottomBar extends StatefulWidget {
     this.interactionGlowRadius = 1.5,
     this.interactionBehavior = GlassInteractionBehavior.full,
     this.pressScale = 1.04,
+    this.indicatorBehindIcons = false,
   })  : assert(tabs.length > 0, 'GlassBottomBar requires at least one tab'),
         assert(
           selectedIndex >= 0 && selectedIndex < tabs.length,
@@ -269,6 +270,16 @@ class GlassBottomBar extends StatefulWidget {
   ///
   /// Defaults to 1.04 (4% growth — matches iOS 26 Apple News pill).
   final double pressScale;
+
+  /// Whether the moving glass indicator is drawn *behind* the tab icons
+  /// instead of *on top* of them.
+  ///
+  /// Defaults to `false`, preserving the original "magic lens" behaviour
+  /// where the translucent indicator refracts the icon underneath. The lens
+  /// effect can wash out saturated icon colours, however — set this to
+  /// `true` for the iOS-native pattern (icons paint over a background pill)
+  /// when icon colour fidelity matters more than the refraction effect.
+  final bool indicatorBehindIcons;
 
   // ===========================================================================
   // Tab Configuration
@@ -533,6 +544,7 @@ class _GlassBottomBarState extends State<GlassBottomBar> {
                 interactionScale: widget.interactionBehavior.hasScale
                     ? widget.pressScale
                     : 1.0,
+                indicatorBehindIcons: widget.indicatorBehindIcons,
                 childUnselected: Row(
                   children: [
                     for (var i = 0; i < widget.tabs.length; i++)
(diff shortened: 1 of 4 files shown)


sdegenaar:
Hey @jfhair,

Thanks so much for taking the time to contribute and for using the package — really appreciate it!

With the glass indicator on top, it refracts the icons and labels through the glass, which is the signature iOS 26 effect. If the icons are moved on top instead, they won't refract, which differs from how iOS 26 bottom bars actually behave.

Do you have a specific iOS 26 Liquid Glass reference image or video you're working from? Would love to see what you're aiming for.

Thanks again! 
Cheers

jfhair:
Hey!  Sorry I didn't see this and respond sooner.  Basically, my issue is that I can't set deeper colors or match any color precisely if the indicator is on top of the icon.  They all end up muted by the indicator.  Here's a screenshot of the bottom bar on my app with the PR and without.  You can see that the colors end up more muted with the indicator on top.  This does not appear to be happening in the apple apps.  I tried to match one of their colors with the indicator on top and it was impossible because of the muting.  Not sure if there's something going on on their side to allow them to stay vivid with an overlay, but this was the best I could come up with to keep the icons from going dull.  Thanks for the response.  Let me know what you think...

IMG_5707
IMG_5706

jfhair:
FWIW, if there's some way that I'm missing to keep the visual effect that the indicator gives while not dulling all of the colors, I'd love that!  Let me know if I'm just missing out on something I need to do.  Thanks!

sdegenaar:
Thanks @jfhair for clarifying. Looking at it now. Yes definitely looks much better and clearer. Good pickup. Actually this looks more like a bug and should be the default behaviour but evaluating it now. Thanks!!

sdegenaar:
Hey @jfhair — really appreciate you taking the time to dig into this and put together a proper PR. This was a legitimate bug and your diagnosis was spot on.

We couldn't merge the PR as-is because moving the indicator fully behind the icons breaks the refraction effect — the glass shader needs icon content beneath it to warp the pixels as the pill moves. That's what made this tricky.

The fix we landed in 0.9.3 uses a split-pass approach: the pill's solid background renders below the icons (so they're vibrant at rest), while the glass shader still renders above them (so refraction is preserved during animation). Your PR was the direct trigger for working this out.

You're credited by name in the [0.9.3 CHANGELOG](https://github.com/sdegenaar/liquid_glass_widgets/blob/main/CHANGELOG.md) — thanks again for the sharp eye and for contributing. 🙏

Cheers

PR #37, opened 2026-04-30
feat(shader): asymmetric corner radius + clip geometry fix for glass sheets

yukinoaruu:
### Background

Previously the lightweight glass shader had no proper support for
asymmetric shapes: all corners, edges, and rims were computed assuming
a single uniform radius. Because of this, GlassModalSheet and GlassSheet
exposed a `forceSpecularRim` parameter as a workaround — it bypassed the
shader's rim/edge rendering entirely and instead drew specular highlights
and edges manually on a Canvas layer. This produced acceptable results,
but required an explicit opt-in and could not correctly handle asymmetric
geometry (different top and bottom radii).

With this PR the package core handles asymmetric shapes correctly at the
shader level. `forceSpecularRim` is no longer required for proper edge and
corner rendering and can be left at its default (`false`).

### Problem

Two related rendering defects existed on Skia/Impeller (GlassQuality.standard, GlassQuality.standard.premium):

**1. Wrong corner radii for GlassModalSheet**
The lightweight glass shader used a single uCornerRadius float for all
four corners. GlassModalSheet renders via LiquidVerticalRoundedSuperellipse,
which has distinct top and bottom radii. The shader had no way to express
this — all corners were rendered with a hardcoded fallback of 16px,
producing a visually incorrect shape.

**2. Transparent gaps near corners**
The ClipPath in LightweightLiquidGlass used a superellipse
(RoundedSuperellipseBorder) while the shader SDF draws a circular-arc
rounded rectangle. A superellipse extends further into corners than a
circular arc at the same radius, meaning pixels in that zone were:
- preserved by the ClipPath (inside the superellipse)
- made transparent by the shader (outside the circular arc)
This produced small but visible transparent notches at corner transitions
in both GlassModalSheet and GlassSheet.

### Solution

**Asymmetric shader mode**
Extend the shader with a per-corner radius mode gated by a sentinel:
- uCornerRadius = -1.0 → asymmetric mode; per-corner radii are read
  from the new uData6 uniform (slots 24-27): (topLeft, topRight,
  bottomRight, bottomLeft)
- uCornerRadius >= 0 → existing symmetric path, zero behavioural change
On the Dart side, _updateShaderUniforms detects
LiquidVerticalRoundedSuperellipse by static type and writes the
sentinel + four corner radii. All other shapes follow the original path
and write 0,0,0,0 for uData6 (ignored by the shader).

**ClipPath geometry alignment**
Replace the superellipse ClipPath with a matching circular-arc
RoundedRectangleBorder for both superellipse shape types, so the
clip boundary aligns exactly with the shader SDF boundary:
- LiquidVerticalRoundedSuperellipse → BorderRadius.vertical
- LiquidRoundedSuperellipse → BorderRadius.all

### Scope

- Affects only the lightweight shader path (Skia/Web).
  Impeller/premium path is completely untouched.
- No API changes, no new parameters, no breaking changes.
- forceSpecularRim remains available for backwards compatibility,
  but is no longer needed for correct rendering.

### Files changed

shaders/lightweight_glass.frag
  - Add uData6 uniform (slots 24-27)
  - Quadrant-dispatched SDF when uCornerRadius < 0

lib/widgets/shared/lightweight_liquid_glass.dart
  - Asymmetric mode: populate uData6 for LiquidVerticalRoundedSuperellipse
  - Symmetric shapes: unchanged code path + zero-filled uData6
  - ClipPath fix: use circular-arc RoundedRectangleBorder for
    LiquidVerticalRoundedSuperellipse and LiquidRoundedSuperellipse

### As it was before with the forceSpecularRim parameter disabled:

изображение

### What it looks like now after finalizing the work of shaders with asymmetric widgets and the forceSpecularRim parameter turned off:

изображение

```dart
void _showChameleon(BuildContext context) {
    GlassModalSheet.show(
      context: context,
      enablePeek: true,
      fillTransition: FillTransition.gradual,
      quality: GlassQuality.standard,
      barrierColor: Colors.black45,
      enableSaturationGlow: false,
      enableInteractionGlow: false,
      forceSpecularRim: false,
      builder: (ctx) => const ChameleonContent(),
    );
  }
```

### Floating Peek (Floating Pill):
**Implemented support for the "floating" state via new parameters: peekWidth, peekHorizontalMargin, and peekBottomMargin.**
The sheet can now appear as a compact, independent "pill" at the bottom of the screen without touching the edges.
Geometric Morphing Engine:

**Introduced non-linear metrics interpolation in _calculateMetrics.**
Snappy Feel: The transition from the peek state to half is compressed into the first 15% of movement (threshold: 0.15). This creates the signature "springy" iOS effect, where a small pill rapidly transforms into a full-width panel.
Adaptive Radius Interpolation:

**Added peekTopBorderRadius and peekBottomRadius.**
The system now successfully morphs a "rounded pill" into a rectangular sheet with clean, refined top corners during expansion.
Tactile Calibration:

**Configured Curves.easeOut for geometric parameters, providing a sense of visual "lightness" and high responsiveness during dragging.**

### Demo Apple Style Peek:

https://github.com/user-attachments/assets/e16d7f16-fffa-46df-820b-8df5280afea1

```dart
void _showMapsStyleSheet(BuildContext context) {
    GlassModalSheet.show(
      context: context,
      peekSize: 90.0,
      peekWidth: 340.0,
      peekHorizontalMargin: 20.0,
      peekBottomMargin: 32.0,
      peekTopBorderRadius: 30.0,
      peekBottomRadius: 30.0,
      horizontalMargin: 8.0,
      bottomMargin: 8.0,
      topBorderRadius: 64.0,
      initialState: SheetState.peek,
      enablePeek: true,
      mode: SheetMode.persistent,
      quality: GlassQuality.standard,
      forceSpecularRim: false,
      builder: (ctx) => const MapsStyleContent(),
    );
  }
```


sdegenaar:
Hey @yukinoaruu  — just wanted to say a huge thank you for this contribution! 

The asymmetric corner radii work is now shipping in v0.9.5 and the implementation is really clean — the sentinel approach keeps the symmetric path completely unchanged, and the shader compiles correctly across all platforms including Windows/SkSL. Great attention to detail.

If you have an example app or demo showcasing this all we'd love to see it! From your videos looks really useful. Feel free to open a PR to add it to the example folder or send to me whatever is easiest. It's a great feature and more real-world examples would help the community understand how to use it.

Thanks again — really appreciate you taking the time to contribute!

Cheers

PR #45, opened 2026-05-03
Fix: snap-to-state side effects skipped after drag-resolved transitions

jfhair:
## Summary

`_GlassModalSheetState._snapToState` decides whether to fire its side-effects branch (haptics, `onStateChanged`, `_scrollController.jumpTo(0)` on collapse) by comparing the requested target against `_currentState`. But `_currentState` is mutated **during drag** by `_applyDrag` and `_jumpTo` — they call `_geometry.resolveTarget` on the in-flight position and silently commit the result so the in-flight visuals (radii interpolation, expand-progress, glass settings) can interpolate toward the resolved snap target.

That mid-gesture mutation defeats the equality check: if the drag itself crosses a snap threshold, by the time the user releases and `_onPointerUp` calls `_snapToState(target)`, `_currentState` already equals `target` — and the entire side-effects branch is silently skipped.

## Symptoms

- **No haptic** on slow-drag settle. Fast snaps still haptic'd because the gesture didn't linger long enough for `_applyDrag` to commit the target.
- **`onStateChanged` never fires** for slow-drag transitions — the consumer's mirrored state field stays stale, and any state-derived UI (e.g. switching content scroll physics between full and half) never updates.
- **`_scrollController.jumpTo(0)` never runs** on a slow-drag full → half collapse, so a `SingleChildScrollView` wired to the package's `ScrollControllerProvider` keeps its full-state scroll offset into the half-state viewport (header cropped above the visible area, no way to recover because half-state physics is `NeverScrollable`).

## Repro

Drag a sheet slowly from full to half (slowly enough that the drag spans several frames during which the resolved snap target crosses `half`). Watch:
- The mirrored state field passed via `onStateChanged` stays at `full`.
- No haptic on settle.
- Any state-derived UI in the sheet content stays in its full-state mode.

The included test (`onStateChanged fires after slow drag past snap threshold`) reproduces this with `tester.startGesture` + 20× `moveBy(15)` calls spaced 16ms apart. Fails on `main`, passes under this change.

## Fix

Track the published state in a separate `_settledState` field, initialized from `widget.initialState` and only updated inside `_snapToState`'s side-effects branch. The equality check now correctly answers _\"did the state change since the last time we told consumers about it\"_ regardless of how many times `_applyDrag` reshuffled `_currentState` during the gesture. `_currentState` keeps its existing semantics for in-flight visuals.

Single-line conceptual change in `_snapToState` (`state != _currentState` → `state != _settledState`, plus assigning `_settledState` alongside `_currentState` inside the branch). New field, new initState assignment, doc comment, and a regression test.

## Test plan

- [x] All 159 existing tests still pass.
- [x] New regression test (\"onStateChanged fires after slow drag past snap threshold\") passes under this change and fails on `main`.

### DIFF
--- a/lib/widgets/overlays/shared/glass_modal_sheet_state.dart
+++ b/lib/widgets/overlays/shared/glass_modal_sheet_state.dart
@@ -12,6 +12,27 @@ class _GlassModalSheetState extends State<GlassModalSheet>
   SheetState get _currentState => _currentStateNotifier.value;
   set _currentState(SheetState v) => _currentStateNotifier.value = v;
 
+  /// The state most recently published to consumers via
+  /// [GlassModalSheet.onStateChanged] (and through which any
+  /// snap-to-state side effects — haptics, scroll-to-top — fired).
+  ///
+  /// Distinct from [_currentState]: that value is mutated during
+  /// drag by [_applyDrag] and [_jumpTo] to drive in-flight visual
+  /// interpolation (radii, expand-progress, glass settings) toward
+  /// the resolved snap target. Comparing the snap target to
+  /// [_currentState] in [_snapToState] therefore misses the
+  /// transition whenever the drag had already crossed a snap
+  /// threshold mid-gesture: by the time the user releases and
+  /// [_onPointerUp] calls `_snapToState(target)`, [_currentState]
+  /// already equals `target`, and the side-effects branch is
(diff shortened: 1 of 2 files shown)


sdegenaar:
Incorporated in d3fd249. The _settledState fix is exactly right, and your regression test is included. Will be release in next version v0.10.1. Thanks again @jfhair! 🙌

PR #52, opened 2026-05-07
TabBar: Fix dragging, constraints, and z-order; add Divider/Shadow settings

g3mf0r:
Fix: indicator responsiveness and boundary constraints
- Indicator Lag: Replaced delta-based dragging with absolute pointer positioning to eliminate drift in non-scrollable mode
- Сorrected z-order in scrollable mode to prevent indicator from overlapping text
- Boundary Clamping: Added constraints to prevent the indicator from flying outside the bar's visible bounds

Feats:
- Customizable Dividers: Added DividerSettings (thickness, indents, decoration) with auto-hide logic for the active tab
- Custom Shadows: Added support for a list of indicatorShadow for the glass pill
- Grab-to-Drag: Refined gesture logic to allow the active tab to be grabbed and dragged immediately, even in scrollable mode
- Gesture Priority: Integrated GestureArenaTeam to let the indicator steal focus from the scroll view

I could continue working on it, but I didn't touch it since it's still a tab bar. If you could move these features to SegmentedControl (I could add behavior for both modes, as well as smaller things)

https://github.com/user-attachments/assets/4dda3f1d-5b82-4af6-88ae-60a5d16ed63a

### DIFF
--- a/lib/widgets/shared/animated_glass_indicator.dart
+++ b/lib/widgets/shared/animated_glass_indicator.dart
@@ -68,6 +68,9 @@ class AnimatedGlassIndicator extends StatelessWidget {
   /// Used in scrollable mode where tabs have different widths.
   final double? exactOffset;
 
+  /// Optional list of shadows to be cast by the indicator background.
+  final List<BoxShadow>? shadows;
+
   const AnimatedGlassIndicator({
     super.key,
     required this.velocity,
@@ -87,6 +90,7 @@ class AnimatedGlassIndicator extends StatelessWidget {
     this.paintGlass = true,
     this.exactWidth,
     this.exactOffset,
+    this.shadows,
   });
 
   static const _baseGlassSettings = LiquidGlassSettings(
@@ -143,6 +147,7 @@ class AnimatedGlassIndicator extends StatelessWidget {
           decoration: BoxDecoration(
             color: indicatorColor,
             borderRadius: BorderRadius.circular(borderRadius),
+            boxShadow: shadows,
           ),
           child: const SizedBox.expand(),
         ),
--- a/lib/widgets/surfaces/glass_tab_bar.dart
+++ b/lib/widgets/surfaces/glass_tab_bar.dart
@@ -120,6 +120,8 @@ class GlassTabBar extends StatefulWidget {
     this.indicatorSettings,
     this.backgroundKey,
     this.maskingQuality = MaskingQuality.high,
+    this.dividerSettings,
+    this.indicatorShadow,
   })  : assert(tabs.length >= 2, 'GlassTabBar requires at least 2 tabs'),
         assert(
           selectedIndex >= 0 && selectedIndex < tabs.length,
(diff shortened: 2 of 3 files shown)


sdegenaar:
Hey @g3mf0r  — first of all, really impressive work here. The gesture arena approach for the indicator drag, the absolute-coordinate fix for the drift on desktop/web, and the z-order correction are all exactly the kind of quality contributions this package needs. Thank you. 🙏

We're still in final review before merging and wanted to have a quick discussion with you about two of the new features before we do.

indicatorShadow We noticed this applies a BoxShadow to the solid background pill — the resting-state selection indicator underneath the glass layer. We want to make sure we're not missing something before we merge it: have you seen Apple use a drop shadow on the tab selection indicator anywhere in iOS 26? From what we can see in the native tab bar, depth comes from the glass contrast and bar material rather than a traditional shadow. We're wondering if there's a reference or use case we've overlooked that made you add this?

DividerSettings Similar question — we love the idea of giving developers more control, but we want to check: did you see dividers used between tab items anywhere in Apple's iOS 26 design, or was this more of a general customisation you thought would be useful? Our concern is that dividers feel more Material Design than Liquid Glass, and we want to be careful about introducing elements that stray from the native aesthetic.

Totally possible we're being too strict here and you have context we don't — that's why we wanted to ask before making any decisions. If you can point us to any references or explain the thinking, we're very open to it. And if after discussion you agree with the concern, would you be willing to update the PR and we'll get it merged straight away?

Either way, the gesture work is going in — just want to get the design side right first. 🙌

Cheers

g3mf0r:
No, I haven't seen anything like this with Liquid Glass. At least, I don't think I've seen many things they've done that I can achieve with this package, unlike regular Liquid Glass. I also don't think I've seen anything with Apple that allows for scrolling, each segment to be a different size, or that allows me to round the corners. You should probably point out if this package relies solely on their guidelines, but I think it would be nice to have the option for everyone to decide whether to deviate from this and customize it, and let the default values ​​completely replicate the behavior as Apple intended

When adding this, I was guided by the behavior of the SlidingSegmentedControl widget

Again, I think these changes belong in SegmentedControl. It was just easier for me to implement it here because Tabbar already had everything that SegmentedControl did. To imitate SegmentedControl, it was enough to create a different BorderRadius. Besides, as far as I understand, SegmentedControl is still being finalized by you; I didn't find any code in shared like Tabbar

Regarding

> Our concern is that dividers feel more Material Design than Liquid Glass

You can configure this 1:1 like a standard SegmentedControl. To be honest, I've never even seen SegmentedControl in Material Design

sdegenaar:
Thanks. This discussion is good.

You're right — GlassSegmentedControl doesn't have the same shared structure as GlassTabBar yet. I am going to refactor it to match. Once that's done, would you be interested in adding DividerSettings there in a follow-up PR? It would be the right home for it and the architecture would be clean. 

What do you think if I just merge the gesture fixes from this PR plus the refactor of GlassSegmentedControl?

g3mf0r:
Yes, of course, you can change this at your discretion, remove the dividers and shadows, I would like to see them in SegmentedControl along with the scrollable mode, once you complete it, in general, if you see a different implementation for the tabbar behavior, you can completely reject this PR, although it does not affect the behavior in normal mode, but then you will have to fix a couple of problems yourself, namely, the fact that the fill overlaps the text in scrollable mode, and also the fact that in normal mode the active indicator does not keep up with the finger movement

sdegenaar:
OK good points. I have merged all changes in, as see in light mode it can add to contrast which can be handy. I added some tests as well and refactored GlassSegmentedControl's internals into seperate file to match same pattern used elsewhere. You can add changes in here in follow up PR if you like to match.

Your name is credited in CHANGELOG.md for v0.10.8 which is live on pub.dev. Thanks again — great PR and hopefully more to come  🙏

Any problems let me know,
Cheers

PR #55, opened 2026-05-10
feat(Menu): dynamic screen adjustment and enhanced geometry; fix(TabBar): glass refraction & fixed drag

g3mf0r:
Menu
- Screen Padding: Migrated to EdgeInsets for per-side control and updated autoAdjustToScreen to respect these boundaries
- Smart Alignment: Integrated GlassMenuAlignment enum with auto-flip logic to prevent vertical viewport overflow
- Selection: Added itemBorderRadius
- Morphing Fix: Included calculated offsets in the transformation to ensure correct sliding within screen bounds

TabBar
- Refraction Layer: Added a background layer in non-scrollable mode. Without it, the labels didn't interact with the glass; now the indicator correctly refracts the tabs in active mode
- Fixed Mode Jumping: Fixed a bug where dragging from the first to the last tab snapped the indicator back to the previous tab. It now correctly calculates multi-tab jumps while keeping the original drag sensitivity

### DIFF
--- a/lib/widgets/overlays/glass_menu.dart
+++ b/lib/widgets/overlays/glass_menu.dart
@@ -61,13 +61,24 @@ class GlassMenu extends StatefulWidget {
   /// Typically contains [GlassMenuItem] and [GlassMenuDivider].
   final List<Widget> items;
 
+  /// The alignment of the menu relative to the trigger.
+  final GlassMenuAlignment? menuAlignment;
+
+  /// Whether to automatically adjust the menu position to keep it on screen.
+  final bool autoAdjustToScreen;
+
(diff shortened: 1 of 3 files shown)


sdegenaar:
Hey @g3mf0r, really appreciate this contribution! The new GlassMenuAlignment, padding options, and the multi-tab drag fix are all solid additions.

I've integrated everything into main as part of the v0.10.10 release (commit 466ba8e). I made a small tweak to move the GlassMenuAlignment enum into the public API so it's accessible, and added the test coverage for the new features.

Thanks again for sticking with the project and continuing to improve it! 🙏

Cheers

PR #61, opened 2026-05-18
fix: eliminate rectangular blur halo around rounded glass over iOS PlatformViews

jfhair:
## Summary

When a rounded `liquid_glass_widgets` surface is stacked over an iOS PlatformView (e.g. `mapbox_maps_flutter` `MapWidget`, `video_player` on iOS), a faint **rectangular halo** appears extending past the rounded corners of the glass shape. It's most visible when light backdrop content scrolls underneath — the blurred light pixels bleed into the corner regions outside the rounded silhouette.

Root cause: `LightweightLiquidGlass` and `_FrostedFallback` wrap their glass surface in `ClipPath(ShapeBorderClipper(...))`. Flutter framework [PR flutter/flutter#177551](https://github.com/flutter/flutter/pull/177551) (merged Dec 2025, shipped in 3.41.0-0.0.pre+) fixed this at the engine level by forwarding clip data to the iOS PlatformView mutator stack — but **only for `ClipRRect`, not `ClipPath`**, even when the path inside is mathematically a rounded rect.

This PR routes shapes that resolve to a `RoundedRectangleBorder` (`LiquidRoundedSuperellipse`, `LiquidVerticalRoundedSuperellipse`) through `ClipRRect` instead of `ClipPath`. The engine fix triggers and the halo disappears.

Closes upstream Flutter [#175048](https://github.com/flutter/flutter/issues/175048) and [#115926](https://github.com/flutter/flutter/issues/115926) for `liquid_glass_widgets` users. Also matches the symptom in closed issue [#10](https://github.com/sdegenaar/liquid_glass_widgets/issues/10) of this repo.

## Changes

- **`lib/widgets/shared/lightweight_liquid_glass.dart`** — `LightweightLiquidGlass.build`: when the resolved `clipShape` is a `RoundedRectangleBorder` with a `BorderRadius`, wrap the `_LightweightGlassEffect` in `ClipRRect(borderRadius: ...)` instead of `ClipPath(clipper: ShapeBorderClipper(...))`. Other shapes fall back to `ClipPath` unchanged.
- **`lib/widgets/shared/adaptive_glass.dart`** — `_FrostedFallback.build`: same swap at the two `ClipPath(ShapeBorderClipper(...))` sites (the blur-body wrap and the child-content wrap). Logic extracted into a private `_ShapeClip` helper widget for reuse and clear documentation.
[...]

PR #62, opened 2026-05-18
feat: expose baseAlphaMultiplier / edgeAlphaMultiplier on GlassSearchableBottomBar to control indicator rim floor

jfhair:
## Summary

The interactive indicator shader (`shaders/interactive_indicator.frag`) has a **hardcoded `borderMask * 0.9` alpha floor** at line 274 that overrides any `uBaseAlphaMultiplier` / `uEdgeAlphaMultiplier` the caller passes — ensuring "structural rim even at rest." That floor manifests as a visible **border flash at the end of the morphing-pill animation** when the bar runs at `GlassQuality.standard` over a PlatformView: the lightweight pipeline can't fade the indicator's edge cleanly, so the hardcoded floor becomes the most visible thing during the morph.

There's no way for an app-level caller to kill the floor — the multipliers exist but are clamped out by the `* 0.9` term.

This PR makes the floor scaleable and plumbs the existing internal `baseAlphaMultiplier` / `edgeAlphaMultiplier` knobs out to the public `GlassSearchableBottomBar` API so callers who hit the rim-flash issue can fully suppress it on a per-screen basis without touching package internals.

## Changes

- **`shaders/interactive_indicator.frag`** — scale the hardcoded `borderMask * 0.9` floor by `uEdgeAlphaMultiplier` so callers can fully kill it via `edgeAlphaMultiplier: 0.0`. Default `0.4` preserves the original baseline.
- **`lib/widgets/shared/animated_glass_indicator.dart`** — accept optional `baseAlphaMultiplier` / `edgeAlphaMultiplier`; forward to the `GlassEffect` uniform writer (no behaviour change when null).
- **`lib/widgets/surfaces/shared/searchable_bottom_bar_internal.dart`** — forward the same two optional doubles from the internal indicator widget down to `AnimatedGlassIndicator`.
- **`lib/widgets/surfaces/glass_searchable_bottom_bar.dart`** — public-facing surface: add `baseAlphaMultiplier` / `edgeAlphaMultiplier` to `GlassSearchableBottomBar`'s constructor, with doc comments explaining the rim-flash use case.

## Rationale for exposing these as public knobs

The existing `baseAlphaMultiplier` / `edgeAlphaMultiplier` are already package-internal — this PR just plumbs them out. They felt like the right level of escape hatch for the morphing pill specifically: a caller who hits the rim-flash issue on a busy backdrop (e.g. a live map PlatformView under `GlassQuality.standard`) can set both to `0.0` to render the morphing pill invisible while leaving the bar's own glass surface unchanged.

Happy to discuss an alternative (e.g. a single \`rimVisibility\` enum, or auto-detection of `GlassQuality.standard + PlatformView`) if the API surface feels too low-level.

## Backward compatibility

- **Visual**: pixel-identical with the new defaults (the shader's default is `0.4`, which matches the previous hardcoded floor exactly — `0.4 / 0.4 = 1.0` scaling). No-op for all existing callers.
- **API**: no breakage — all new params are optional (default `null` → package baseline). \`GlassSearchableBottomBar\` constructor gains two new optional named params.

## Test plan

- [x] `flutter analyze` clean on patched files.
- [x] Verified in a real app (over `mapbox_maps_flutter` `MapWidget` on iOS Impeller, `GlassQuality.standard`) — passing `baseAlphaMultiplier: 0.0, edgeAlphaMultiplier: 0.0` fully eliminates the rim flash at the end of the morph; baseline rendering unchanged when both are null.
- [ ] (Draft) Maintainer review of approach + doc-comment wording before publishing.

## Relationship to PR #61

Separate concern from PR #61 (PlatformView halo / `ClipPath`→`ClipRRect`). Different shader, different widget surface. This PR is based directly on `main`, **not** stacked on #61 — either can land first.

### DIFF
--- a/lib/widgets/shared/animated_glass_indicator.dart
+++ b/lib/widgets/shared/animated_glass_indicator.dart
@@ -74,6 +74,22 @@ class AnimatedGlassIndicator extends StatelessWidget {
   /// muddy the liquid glass animation. Pass `null` (default) for no shadow.
   final List<BoxShadow>? shadows;
 
+  /// Forwards to [GlassEffect.baseAlphaMultiplier]. Controls the indicator
+  /// pill's center transparency in the interactive shader. Defaults to
+  /// [GlassEffect]'s default (0.2). Pass 0.0 together with
+  /// [edgeAlphaMultiplier]=0.0 to make the indicator's morphing pill
+  /// invisible (e.g. for screens where the morph-time rim flash isn't
+  /// wanted).
+  final double? baseAlphaMultiplier;
+
+  /// Forwards to [GlassEffect.edgeAlphaMultiplier]. Controls the indicator
+  /// pill's rim opacity in the interactive shader AND scales the
+  /// previously-hardcoded `borderMask * 0.9` floor in
+  /// `interactive_indicator.frag` — so passing 0.0 actually kills the
+  /// rim, instead of running into the floor. Defaults to [GlassEffect]'s
+  /// default (0.4).
+  final double? edgeAlphaMultiplier;
+
   const AnimatedGlassIndicator({
     super.key,
     required this.velocity,
@@ -94,6 +110,8 @@ class AnimatedGlassIndicator extends StatelessWidget {
     this.exactWidth,
     this.exactOffset,
     this.shadows,
+    this.baseAlphaMultiplier,
+    this.edgeAlphaMultiplier,
   });
 
   static const _baseGlassSettings = LiquidGlassSettings(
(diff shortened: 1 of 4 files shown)


sdegenaar:
Thanks for the follow-up on the rim flash, @jfhair. With v0.12.1's ClipRRect routing (from your PR #61) and the uAmbientRim scaling we shipped in 0.12.0, we believe the underlying causes may be addressed.

Could you test against 0.12.1 and let us know if you still see the rim flash? If it persists we can revisit the alpha multiplier approach — but we'd want to apply it consistently across all indicator-bearing surfaces rather than just GlassSearchableBottomBar.

Closing for now — please reopen if the issue remains after 0.12.1.

Thanks again,
Cheers

jfhair:
Tested against v0.12.7 — **no rim flash observed** on our Trail Map use case.

Reproduced the exact scenario the multipliers were originally bandaging:
- `GlassSearchableBottomBar` running at `GlassQuality.standard`
- Sitting over the Mapbox `MapWidget` (iOS PlatformView)
- Bottom-bar morphing between Trail Map ↔ Favorites ↔ Plans tabs

Dropped both `indicatorBaseAlphaMultiplier` and `indicatorEdgeAlphaMultiplier` from our call site on the bump. Standard quality looks visibly different from premium (as expected from the shader path differences), but the morph end-frame is clean — no border flash carryover.

Your hypothesis held: `_ShapeClip` ClipRRect routing from #61 + `uAmbientRim` scaling from 0.12.0 addressed the root cause. Happy to leave #62 closed.

Thanks again — really appreciate the careful follow-through on both fixes.

PR #88, opened 2026-06-06
feat(GlassMenu): GlassMenuController + showDismissBarrier

F1orian:
## Summary

Adds imperative control over `GlassMenu` for cases where something other than the menu's own trigger decides when it appears — e.g. a central gesture arena, a canvas long-press, or a programmatic popover.

### GlassMenuController

New controller class with `open()`, `close()`, and `isOpen`. Pass it to `GlassMenu(controller: ...)`.

Unlike tapping the trigger (which toggles and is gated on morph progress), the controller commands `_openMenu` / `_closeMenu` directly, so a `close()` that lands mid-open reliably dismisses the menu instead of re-opening it.

Lifecycle:
- Attaches in `initState`, detaches in `dispose`
- Hot-swaps correctly in `didUpdateWidget`

### showDismissBarrier

New `bool showDismissBarrier` parameter (default `true`, no breaking change).

Set to `false` when an external gesture owner must keep receiving pointer events while the menu is open and will handle dismissal itself. The default full-screen barrier sits in the root overlay and swallows all input — `showDismissBarrier: false` removes it so the gesture owner stays in control.

Pairs naturally with controller-driven open/close.

## Risk

Both additions are opt-in with defaults that preserve existing behavior. No changes to the trigger-tap path.

### DIFF
--- a/lib/widgets/overlays/glass_menu.dart
+++ b/lib/widgets/overlays/glass_menu.dart
@@ -198,6 +198,15 @@ class GlassMenu extends StatefulWidget {
   /// arena) decides when to show and dismiss the menu.
   final GlassMenuController? controller;
 
+  /// Whether to render the full-screen tap-to-dismiss barrier behind the open
+  /// menu. Defaults to `true` (modal: a tap outside the menu closes it).
+  ///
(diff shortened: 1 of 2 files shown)


sdegenaar:

PR #94, opened 2026-06-09
feat: platformViewBackdrop — premium glass over iOS PlatformViews

jfhair:
## Summary

Adds an opt-in `platformViewBackdrop` flag so **premium-quality glass renders correctly over an iOS PlatformView** (e.g. a Mapbox map or a video), instead of appearing as a solid slab.

## Motivation

The premium shader builds its backdrop from `RenderRepaintBoundary.toImageSync`, and `toImageSync` **cannot capture an iOS PlatformView** — the OS composites PlatformViews outside Flutter's layer tree. So a premium glass surface placed directly over a map renders as an opaque slab (no live view, no blur of the map).

The existing workaround is to drop the whole surface to `GlassQuality.standard`, but that also swaps the widget type (losing premium animations/state) and forfeits the premium look everywhere, not just over the PlatformView.  Trying to toggle between premium and standard quality on a per-page basis does not animate smoothly, requiring multiple workarounds to address.  This approach allows premium to be used app-wide, even when there are one or more app pages with PlatformViews.

## What this does

`platformViewBackdrop` is an **orthogonal axis** to `GlassQuality` — it changes only the *backdrop source*, not the quality tier:

- **`AdaptiveGlass` / `AdaptiveGlass.grouped`** — new `platformViewBackdrop` flag. When `true`, the surface uses the live `BackdropFilter` path **even at `GlassQuality.premium`**, so the PlatformView shows through. Refraction is lost (there's no sampleable texture), but the live blur + rim/lighting remain.
- **`GlassSearchableBottomBar`** — new `platformViewBackdrop` flag, forwarded to the tab indicator and search pill:
  - the bar **background** uses the `BackdropFilter` path;
  - the premium tab **indicator** refracts the bar's own icon layer (a capturable Flutter `RepaintBoundary`) instead of the uncapturable PlatformView region — so the premium magic-lens animation survives over the map with **no quality swap**;
  - the collapsed search button drops to the lightweight veil over a PlatformView so it matches the bar background.

## Backward compatibility

Every flag defaults to `false` — no behaviour change when unset.

## Known limitation (documented on the field)

The premium indicator refracts the icon layer via `toImageSync`, which asserts the captured boundary is clean. While the indicator animates, that layer repaints every frame, so a mid-animation capture can fail and the indicator briefly flashes dark. It's negligible at `magnification` ≈ 1.0 (the default) but grows with magnification — so keep magnification near 1.0 over a PlatformView. (The PlatformView itself can't be captured, so there's no way to refract it directly.)

## Testing

`flutter analyze lib` is clean. Exercised in a production app: a `GlassSearchableBottomBar` over a full-screen Mapbox `MapWidget` — the live map shows through the bar and indicator, premium tab-morph animations play, no solid slab, no quality swap.

## Diff

3 files, +72/−4 — additive only.

### DIFF
--- a/lib/widgets/shared/adaptive_glass.dart
+++ b/lib/widgets/shared/adaptive_glass.dart
@@ -199,11 +199,12 @@ class AdaptiveGlass extends StatelessWidget {
 
     // platformViewBackdrop forces the live BackdropFilter path even at premium:
     // the premium shader's toImageSync backdrop can't capture a PlatformView, so
-    // over one it must use BackdropFilter (live) instead.
+    // over one it must use BackdropFilter (live) instead. The local/cheap checks
+    // are evaluated before the platform shader-support query (_canUseImpeller).
     final bool canUsePremiumShader = !kIsWeb &&
-        _canUseImpeller &&
+        !platformViewBackdrop &&
         quality == GlassQuality.premium &&
-        !platformViewBackdrop;
+        _canUseImpeller;
 
     if (!canUsePremiumShader) {
       // 1. Detect Grouped Elevation
--- a/lib/widgets/shared/adaptive_glass.dart
+++ b/lib/widgets/shared/adaptive_glass.dart
@@ -44,6 +44,7 @@ class AdaptiveGlass extends StatelessWidget {
     this.allowElevation = true,
     this.glowIntensity = 0.0,
     this.isInteractive = false,
+    this.platformViewBackdrop = false,
     super.key,
   });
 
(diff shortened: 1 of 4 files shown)


jfhair:
Hey... just wanted to follow up on my earlier comments about work on improving light mode fidelity.  That's still in progress and getting close, but I've run into a few blockers (like this PR), plus I have a busy week, so it may not be coming imminently.  I did look at your changes in 0.15.1 and they are complementary.  I've already refactored around 0.15.0 and I'll pull in 0.15.1 and test against it as well.  You move fast!  Just wanted you to know that this PR was intentionally narrow and a separate one is still coming.  Thanks as always!

sdegenaar:
Hey @jfhair,
Have merged it it. Thanks so much for this PR. Really appreciate the contribution. Let me know how you go with light mode fidelity in the 0.15.2 release and if there is anything missing and you can help with.

Cheers!

ChibangLW:
Thank you for this PR. I have tested it and can confirm it is no longer a grey blob over platform views. 

Maybe I have misread your description ("... but the live blur + rim/lighting remain."), but I was under the impression, that blur should work somewhat normal over platform views. In my testing the bottom nav bar is completely clear as soon as it is over a platform view while blur is working normally every where else.

jfhair:
Unfortunately, there is no way to implement blur in anything but its most extreme form over an iOS PlatformView.  This is a Flutter limitation which is not likely to be addressed in the near future.  This PR at least allows you to retain most of the look and feel of the premium controls and not have to change quality back and forth when rendering over a PlatformView on one screen and native Flutter content on another.  However, you will need to use some combination of glassColor and opacity to keep the control from being transparent over the PlatformView.  With a bit of tuning, the effect varies between acceptable and quite good, depending on the look you're going for.

PR #100, opened 2026-06-11
feat: whitenStrength — opt-in light-mode whitening (legibility veil) across all quality tiers

jfhair:
## Summary

Adds an opt-in **light-mode whitening ("legibility veil")** to `LiquidGlassSettings` — `whitenStrength` + `whitenGated` — rendered consistently across all three quality tiers, plus an optional **whiten-at-bottom** boost on `GlassSearchableBottomBar`. Both default to no-op (`whitenStrength: 0.0` → byte-identical rendering for existing users).

## Motivation

iOS 26 light-mode glass doesn't just blur and tint — it lays an even whitening layer over the refracted content so glass stays legible over busy light backgrounds, while letting the content remain visible through it. Today the closest approximation in the package is a heavy semi-opaque white `glassColor` (e.g. `Color(0x99FFFFFF)` as in the example app). We A/B-tested that approach against this PR's whitening in a production app (bottom nav over scrolling light content and over a full-screen map) and found the static tint has three drawbacks the whiten avoids:

- it **desaturates/dulls colorful content** behind the glass noticeably more,
- it reads **gray rather than white** (the tint mixes with mid-tone content; whiten lifts toward white *after* the glass render),
- it's **static** — at the bottom of a scroll, where content crowds right up under a bar, there's no extra legibility, which is exactly where iOS lifts hardest.

The new GPU SDF shadows (0.15.2) solve light-mode *elevation*; this PR is the complementary half: light-mode *fill legibility*.

## Visual comparison

Same bar, same light content behind it (iPhone, Impeller, premium tier):

| Gated whiten — this PR (`whitenStrength: 0.30`) | Static white tint (`glassColor: Color(0x99FFFFFF)`) |
|---|---|
| gated whiten | static tint |

The gated whiten reads **white** and preserves the color of the content bleeding through, while dark glyphs stay crisp; the static tint reads **gray** and desaturates everything behind it uniformly.

### `LiquidGlassSettings.whitenStrength` / `whitenGated`

- **`whitenStrength`** (0..1, default **0.0** = off): the finished glass is lifted toward white — `mix(glass, white, whitenStrength)` — as the last step of the render. A single control-wide value, so there are no spatial seams or halo artifacts.
- **`whitenGated`** (default `true`): when gated, the lift is scaled by a per-pixel luminance gate (`smoothstep(0.40, 0.80, luma)`) so **bright content lifts toward white while dark content (text, icons) stays crisp** — the iOS light-mode behaviour. Ungated applies the lift uniformly — useful in dark mode as a subtle frost, where a luminance gate over all-dark content would zero the effect out.

Rendered per tier from the same knob, calibrated so one value reads consistently across tiers:

| Tier | Mechanism |
|---|---|
| premium (Impeller) | fragment shader, slots 18/19 (`uWhiten`, `uWhitenGated`), gate per-pixel |
| standard (lightweight) | glassColor lerped toward white (gain 1.5) before upload |
| minimal (frosted) | same veil applied to the fallback tint |

### `GlassSearchableBottomBar` whiten-at-bottom (opt-in)

`whitenAtBottom` / `whitenBottomThreshold` / `whitenAtBottomTarget` / `scrollController`: when a `scrollController` is provided, the bar animates its effective `whitenStrength` toward `whitenAtBottomTarget` (default 1.0) as the page nears its scroll bottom, and back as it leaves — a 200 ms eased boost. **Inert unless a `scrollController` is passed**, so the defaults change nothing. Light-mode only (resolved via `CupertinoTheme.brightnessOf`); in dark mode only the recipe's base `whitenStrength` applies.

*Design note: the boost lives on the bar (which owns the glass settings and sits over the scrolling content), in its own commit. If you'd rather see this integrated at the `GlassScaffold` / scroll-edge level instead, happy to rework it that way.*

## Backward compatibility

- `whitenStrength` defaults to `0.0`: the shader's whiten mix is a no-op and the veil lerps are skipped — existing rendering is unchanged (full test suite + goldens pass unchanged).
- All new parameters are additive with no-op defaults.

## Testing

- `flutter analyze` clean; full suite passes (the only failures on my host are the pre-existing golden mismatches, byte-identical with and without this change).
- New tests: settings (defaults / `copyWith` / `lerp` / equality), the minimal-tier veil path, and whiten-at-bottom engage/disengage with a live `ScrollController`.
- Exercised in a production app across premium/standard/minimal on-device (iPhone, Impeller): one whiten value reads consistently across tiers; gated whiten preserves dark glyphs over light scrolling content.

## Diff

8 files, +438/−4 — additive only; 2 of the 8 are tests.

### DIFF
--- a/lib/src/renderer/liquid_glass_settings.dart
+++ b/lib/src/renderer/liquid_glass_settings.dart
@@ -27,6 +27,8 @@ class LiquidGlassSettings with EquatableMixin {
     this.standardOpacityMultiplier = 1.0,
     this.shadowElevation = 1.0,
     this.shadow,
+    this.whitenStrength = 0.0,
+    this.whitenGated = true,
   });
 
   /// Creates [LiquidGlassSettings] using Figma-inspired parameter names.
@@ -219,6 +221,36 @@ class LiquidGlassSettings with EquatableMixin {
     return GlassShadow.scaled(shadowElevation);
   }
 
+  /// Light-mode whitening ("legibility veil") amount, from 0 to 1.
+  ///
+  /// On the Premium (Impeller) path this is applied as the last step of the
+  /// render: the finished glass is mixed uniformly toward white,
+  /// `mix(glass, white, whitenStrength)`. Because it is a single control-wide
+  /// value (not per-pixel) there is no spatial seam, so no halo or outline
+  /// artifacts. This models iOS 26 light-mode glass, which lays an even
+  /// whitening layer over the refracted content for legibility on bright
+  /// backgrounds — content stays visible, just whiter.
+  ///
+  /// The Standard and Frosted paths approximate the same knob by lifting the
+  /// glass tint toward white (a "veil"), so a single value reads consistently
+  /// across all quality tiers.
+  ///
+  /// Defaults to 0.0, which disables whitening entirely.
+  final double whitenStrength;
+
+  /// Whether [whitenStrength] is luminance-gated (true, the default) or
+  /// applied uniformly (false).
+  ///
+  /// Gated: the lift only affects brighter pixels and leaves dark content
+  /// (text, icons) crisp — the light-mode behaviour. Ungated: an even lift
+  /// across the whole control — useful in dark mode to give dark glass a
+  /// subtle frost, where a per-pixel gate over an all-dark backdrop would
+  /// zero the whitening out entirely.
+  ///
+  /// Only affects the Premium (Impeller) path; the Standard and Frosted
+  /// approximations are always uniform.
+  final bool whitenGated;
+
   /// The effective saturation taking visibility into account.
   double get effectiveSaturation => 1 + (saturation - 1) * visibility;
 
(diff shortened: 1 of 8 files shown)


jfhair:
Hey... it continues to be a busy week and though I'm hoping to get back to coding tomorrow, you're moving quickly and I wanted to go ahead and get this to you in its current state to be sure you're happy with the direction.  I tried a lot of approaches attempting to match Apple's light mode effects for content under controls, which noticeably whiten while also preserving color, and this is what I landed on.  I like the look, and in combination with whiten-at-bottom, it appears very Apple-like to me.

I attempted to create a similar effect without the premium shader for standard and minimal qualities, but never came up with anything that I was happy with.  I think a simple color/opacity is probably the right approach there (while also using the whiten-at-bottom switch).  However, I do think this looks better when rendering controls in premium quality.  Curious for your thoughts.  Thanks!

sdegenaar:
Merging — this is really solid work @jfhair, nice one 🙏

Don't undersell the Standard/Minimal tiers — the 1.5× gain calibration reads well across all three, that's the hard part done. Premium should look best, that's the whole point of the tier system. The per-pixel luminance gate is something only a fragment shader can do and that's fine.

The whiten-at-bottom is the standout piece — that scroll proximity boost is what makes it feel genuinely Apple-like. Great call on that.

Shader checked out clean for Windows/SkSL too. Will not do anything in this area (light more) so want step on your toes. This is good work and really appreciate the help!!!! Looking forward to your next one! 

Cheers

sdegenaar:
FYI @jfhair — added a whiten slider + side-by-side comparison cards to the Buttons & Shadows demo page so you can A/B the effect in real-time. Also wired up the scrollController + whitenAtBottom on the demo bar so scrolling to the bottom triggers the boost. Feel free to expand it if you want to showcase more and use it to play with.. maybe toggling between the three quality tiers...

Issue #102, opened 2026-06-11
Content-aware light/dark glass adaptation (iOS 26 behavior) for bars/surfaces


isenbj:
On iOS 26, Apple's Liquid Glass continuously evaluates the luminance of the content behind a glass surface and flips the glass between its light and dark variants as content scrolls beneath it. This keeps labels/icons legible no matter what passes under the glass.

liquid_glass_widgets doesn't currently have an equivalent. In our app, GlassSearchableBottomBar (premium quality) floats over a photo grid: the refraction correctly shows the photos through the bar, but the glass tint and label/icon colors are static. When a run of bright photos scrolls behind the bar, contrast drops and the tab labels get hard to read. Mostly fine on dark content; degrades on light content.

jfhair:
We hit exactly this in our app (premium `GlassSearchableBottomBar` over scrolling content) and ended up building a content-aware adapter that I think matches the iOS 26 behavior fairly closely. Happy to turn it into a PR if there's interest — sketching the approach first to get feedback on the shape.

**Decision logic — a contrast vote, not a luminance threshold.** We sample the bar's own rect, grid it into ~6 cells, and for each cell ask the actual question: "which glyph color reads better over this content — the light variant's dark icons, or the dark variant's light icons?" using WCAG contrast ratios (icon luma vs cell luma). Majority across cells picks the bar's `Brightness`; ties keep the current mode (sticky). Because dark glyphs hold contrast over light *and* medium content, the vote naturally keeps the bar light through the medium range and only flips when content is genuinely dark — which matches what Apple does and resists mode-flapping on mixed content like a photo grid.

**Two levers, like the App Store.** Apple darkens the bottom scrim *early* (medium content already pulls it down) but flips the *controls* late. So the adapter emits both: a discrete `Brightness` for the glass + icons, and a continuous 0→1 darkness we animate on a scrim. Deliberately different thresholds.

**Sampling cost.** Today we do a throttled (~180 ms), single-flight, heavily-downscaled async `toImage` of the content boundary and read only the bar band. That works, but for the package I think there's a better split: at premium quality `GlassEffect` already captures the backdrop every frame for refraction, so the luminance vote could piggyback on that existing capture at ~zero extra cost; the explicit sampler would only be needed for standard/minimal (the `BackdropFilter` path has no capture to reuse).

[...]

PR #103, opened 2026-06-12
Content-aware light/dark adaptation: GlassContentAwareScope + bar adaptiveBrightness (#102)

jfhair:
# Content-aware light/dark adaptation — `GlassContentAwareScope` + bar convenience flags

Implements the first iteration of #102, following the shape we agreed on there: the scope + the bar convenience flags. The premium `GlassEffect`-capture piggyback stays a fast-follow, as discussed — the scope sampler is the universal path and lands first.

## What this adds

```dart
GlassContentAwareScope(
  child: Scaffold(
    extendBody: true, // content scrolls underneath the bar
    body: GlassContentAwareContent(
      child: ListView(...),
    ),
    bottomNavigationBar: GlassBottomBar(
      adaptiveBrightness: true,
      onBrightnessChanged: (b) => /* flip your own icon colors */,
      ...
    ),
  ),
)
```

- **`GlassContentAwareScope`** wraps the screen and owns the sampling engine. It captures the content **once** per sample (throttled, heavily downscaled async `toImage` of the content boundary) and serves every registered control from that one image — N controls share 1 capture, and each control evaluates *its own rect* with its own grid, so a small control over a dark photo can flip dark while a wide bar over mixed content stays light, like iOS.
- **`GlassContentAwareContent`** marks the sampled region (a `RepaintBoundary` around the scrolling content). Controls stay outside it, so the capture sees the content *behind* them rather than their own rendering.
- **`GlassContentAwareBrightness`** is the per-control consumer — the bars use it internally, and custom glass controls can wrap themselves in it directly (`builder` receives the committed `Brightness` plus the animated `darkAmount`).
- **`adaptiveBrightness: true`** on `GlassBottomBar` / `GlassSearchableBottomBar` auto-swaps the `GlassTheme.of(context)` light/dark variants (settings, glow palette, radii) and the Cupertino label-derived icon/label colors, with an **`onBrightnessChanged`** callback for everything the package can't see.
- **`brightnessOverride: ValueListenable<Brightness>`** bypasses the sampler entirely — the PlatformView escape hatch from #94 (e.g. key it off the active map style).

## Decision logic — a contrast vote, not a luminance threshold

Each control's rect is gridded (6×1 for bars); every cell votes for whichever glyph color reads better over the cell's average content — the light variant's dark glyphs vs the dark variant's light glyphs — using WCAG contrast ratios. Because dark glyphs hold contrast over light *and* medium content, the vote keeps controls light through the medium range and only flips when content is genuinely dark, which matches the native behavior and resists mode-flapping on mixed content like photo grids. Unpainted pixels are evaluated as a configurable `backgroundColor` so transparent scaffolds vote as what the user actually sees.

**Two levers, like the App Store — this PR ships the first.** The verdict here is the discrete lever (glass + glyphs, flips late). For the transition itself, `GlassContentAwareBrightness.builder` exposes `darkAmount` — the animated 0→1 fade position — so apps can fade their own elements in lockstep with a flip. The *independent* continuous lever from the issue thread (a scrim-darkness axis driven by mean content luminance with earlier cutoffs, so a bottom scrim darkens over medium content while the controls stay light) is deliberately **not** in this iteration: it wants a content-luminance output on the scope (the engine already averages cell luminance for the vote — it just isn't exposed), and the natural consumer is `GlassScrollEdgeEffect`, the surface `GlassScaffold` already renders under the bars. Scrim band registered as its own client + exposed luminance + an adaptive mode on the edge effect = the full App Store behavior, entirely within existing package surfaces — so it slots cleanly into the fast-follow alongside the premium `GlassEffect`-capture piggyback. Apps that already render their own scrim can keep driving it from their own signal in the meantime.

## Flap prevention, animation, throttling — per your review notes

- **Sticky ties + dual-threshold hysteresis.** A tied or sub-threshold vote keeps the current appearance, and the two directions use separate cutoffs (`lightToDarkThreshold` / `darkToLightThreshold`, vote fractions in `[0.5, 1.0]`, default 0.6), so content sitting right at the contrast boundary cannot oscillate.
- **Animated flips.** Verdict changes cross-fade over `flipDuration` / `flipCurve` (default 200 ms ease-in-out, matching the iOS 26 transition). Continuous theme values lerp throughout via the new `GlassThemeVariant.lerp` (null-aware: partial-override fields that exist on only one side switch at the fade midpoint instead of flashing through zero); the discrete residue (light-mode shadow gating, dynamic-color resolution in user content) flips at the midpoint, where the blend hides it.
- **Scroll-aware throttle.** A `ScrollNotification` listener in the scope starts a periodic sampler (`sampleInterval`, default 180 ms) on scroll start and stops it on scroll end with one trailing sample for the settled state; `ScrollMetricsNotification` covers idle content changes. An idle screen samples zero times. Captures are single-flight and downscaled to ~16 px per grid cell, so a readback is a few KB regardless of screen size.
- **Naming** — `GlassContentAwareScope`, steering clear of the existing `GlassAdaptiveScope` (quality degradation).

## Compatibility

Defaults are inert: `adaptiveBrightness: false` with no override mounts none of the adaptive machinery, and the bars' classic color resolution is byte-for-byte the historical behavior. Explicit `selectedIconColor` / `unselectedIconColor` still override everything.

## Verification

- `flutter analyze` clean.
- Full CI-equivalent suite (excl. golden/renderer): **1907 tests passing**, including 58 new ones — verdict math units (vote, ties, hysteresis, transparency substitution), the real capture pipeline end-to-end over dark/light/medium/mixed content, the cross-fade + override layers, the scroll-driven sampler lifecycle, registration/teardown edge cases, and both bars' integration (probes inside the bar subtree verify the `MediaQuery`/`CupertinoTheme` swap actually reaches the internals).
- New lib code is at 100 % line coverage locally (`lcov`); golden/renderer failure set on a macOS host verified byte-identical to `main`.
- The approach itself is the productization of the adapter we've been running in production in our app (the App Store-style screens from the #102 thread) — validated on-device in both modes with no measurable jank at scroll-time sample rates.
- This exact implementation was also exercised on-device before submitting (iPhone 16 Pro Max, premium tier, ported into our app's glass test harness over panning/zooming map content): verdict flips in both directions through the real capture pipeline, the ~200 ms cross-fade reads correctly (caught mid-lerp on screen), medium-luminance content stays in the light appearance under continuous interaction-driven sampling, and the session log shows zero exceptions.
- The profile-mode leg of that pass caught a release-only bug that no debug-mode test can see: the sampler's mid-paint guard read `debugNeedsPaint`, whose backing value is only assigned inside an assert — in profile/release it throws on read, and the defensive catch swallowed it, silently disabling the sampler. The guard now lives behind an assert closure (release builds sample post-frame, where the boundary is always clean) with a regression test documenting why it must stay that way.

Happy to adjust naming/defaults wherever you'd like — and the `GlassEffect` capture piggyback for premium is queued as the fast-follow.

### DIFF
--- a/lib/liquid_glass_widgets.dart
+++ b/lib/liquid_glass_widgets.dart
@@ -41,6 +41,7 @@ export 'widgets/shared/adaptive_liquid_glass_layer.dart';
 export 'widgets/shared/glass_accessibility_scope.dart'; // GlassAccessibilityScope + GlassAccessibilityData
 export 'widgets/shared/glass_adaptive_scope.dart'; // GlassAdaptiveScope + GlassAdaptiveScopeData + GlassAdaptiveDiagnostic
 export 'widgets/shared/glass_backdrop_scope.dart'; // GlassBackdropScope — per-screen backdrop isolation
+export 'widgets/shared/glass_content_aware_scope.dart'; // GlassContentAwareScope + GlassContentAwareContent + GlassContentAwareBrightness
 export 'widgets/shared/glass_page.dart' show GlassPage, GlassStatusBarStyle;
 export 'widgets/shared/glass_motion_scope.dart';
 export 'widgets/shared/glass_scroll_edge_effect.dart';
--- a/lib/theme/glass_theme_data.dart
+++ b/lib/theme/glass_theme_data.dart
@@ -1,3 +1,5 @@
+import 'dart:ui' show lerpDouble;
+
 import 'package:flutter/material.dart';
 import '../src/renderer/liquid_glass_renderer.dart';
 import 'glass_interaction_settings.dart';
@@ -96,6 +98,45 @@ class GlassGlowColors {
       glowSpreadRadius: glowSpreadRadius ?? this.glowSpreadRadius,
       glowOpacity: glowOpacity ?? this.glowOpacity,
     );
+  }
+
+  /// Linearly interpolates between two glow palettes.
+  ///
+  /// Color fields that are non-null on both sides interpolate smoothly via
+  /// [Color.lerp]; a field that is null on either side switches discretely at
+  /// the midpoint, because `null` means "resolve a brightness-aware default
+  /// at runtime" (see [GlassThemeData.glowColorsFor]) — fading through
+  /// transparent would misrepresent that. The appearance scalars
+  /// ([glowBlurRadius], [glowSpreadRadius], [glowOpacity]) always
+  /// interpolate smoothly.
+  ///
+  /// Returns null when both [a] and [b] are null. Used by
+  /// [GlassThemeVariant.lerp] to cross-fade between light and dark theme
+  /// variants during content-aware brightness flips.
+  static GlassGlowColors? lerp(
+    GlassGlowColors? a,
+    GlassGlowColors? b,
+    double t,
+  ) {
+    if (identical(a, b)) return a;
+    if (a == null || b == null) return t < 0.5 ? a : b;
+    return GlassGlowColors(
+      primary: _lerpColorField(a.primary, b.primary, t),
+      secondary: _lerpColorField(a.secondary, b.secondary, t),
+      success: _lerpColorField(a.success, b.success, t),
+      warning: _lerpColorField(a.warning, b.warning, t),
+      danger: _lerpColorField(a.danger, b.danger, t),
+      info: _lerpColorField(a.info, b.info, t),
+      glowBlurRadius: lerpDouble(a.glowBlurRadius, b.glowBlurRadius, t)!,
+      glowSpreadRadius:
+          lerpDouble(a.glowSpreadRadius, b.glowSpreadRadius, t)!,
+      glowOpacity: lerpDouble(a.glowOpacity, b.glowOpacity, t)!,
+    );
+  }
+
+  static Color? _lerpColorField(Color? a, Color? b, double t) {
+    if (a != null && b != null) return Color.lerp(a, b, t);
+    return t < 0.5 ? a : b;
   }
 
   /// Fallback glow colors with sensible defaults.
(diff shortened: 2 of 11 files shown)


sdegenaar:
Hey @jfhair,
Merging this in — really great work!!. This is exactly the shape we discussed in #102. The scope-per-screen model with N controls off one capture, WCAG contrast vote instead of a raw luminance threshold, the hysteresis for flap prevention — all spot on. 58 tests with 100% line coverage on new code, well done.

Few things we'll sort on our side:

- There's a pre-existing bug on main where GlassThemeVariant.== and hashCode are missing borderRadius — nothing you did, your lerp just surfaced it. We'll fix that straight away.
- Will tidy the bare catch in _sample() to log in debug mode and clean up a redundant setState — minor stuff.

Keen for the fast-follow — the scrim-band luminance lever and the GlassEffect-capture piggyback. The engine's already got the bones for it. Great foundation.

Cheers

sdegenaar:
FYI merged and shipped as 0.16.0 — now live on pub.dev

Again great work on this @jfhair. I built on your engine with a couple of additions:

GlassScaffold.contentAwareBrightness — the scaffold now handles all the scope/content wiring internally. Users get a one-flag opt-in instead of manually assembling three widgets:

```dart
GlassScaffold(
  contentAwareBrightness: true,
  bottomBar: GlassBottomBar(
    adaptiveBrightness: true,
    ...
  ),
  body: CustomScrollView(...),
)
```

The standalone widgets (GlassContentAwareScope, GlassContentAwareContent) remain available for custom layouts.

Dedicated demo — content_aware_brightness_demo.dart in the Examples tab. Alternating light/dark content bands that force visible bar flips during scrolling. Run it with:

cd example && flutter run -t lib/demos/content_aware_brightness_demo.dart

Feel free to modify, add to this demo as you see fit as a good showcase and visual test harness.

Also fixed a couple of polish items along the way (borderRadius in ==/hashCode, debug error reporting in _sample(), redundant setState in the brightness consumer). All documented in the [changelog](https://github.com/sdegenaar/liquid_glass_widgets/blob/main/CHANGELOG.md#0160).

Really appreciate this addition and your contribution with this!!!!! 

Cheers

jfhair:
Hey @sdegenaar, a design check before we finalize the fast-follow...

We're making good progress on the scrim-luminance lever and have a couple more pieces staged, but as we go to consolidate the bits we've built as app-side one-offs and decide what's worth upstreaming, two questions came up that shape the plan — and whether a couple of our queued PRs are even the right approach. We're holding them until we hear from you.

**1. Achromatic tint: flat vs see-through.** `applyGlassColor` sends low-chroma (white/grey/black) tints down the flat/direct-mix path so white glass frosts toward white. We're replicating Apple's glass — only ever light/dark, always see-through — so a near-neutral tint on the flat path reads as a flat slab for us. We've been carrying a local pin (`chromaWeight = 1.0`) to force the luminosity path, and we get the white-frost look from `whitenStrength` instead.

Now that `whitenStrength` exists: would you prefer defaulting achromatic tints to the luminosity (see-through) path, or would you rather expose a small flag for it? We have a PR staged that adds an explicit `GlassTintBlend` (auto/luminosity/flat) enum — but if you'd rather flip the default (or consider the flat-achromatic path legacy), that's simpler than new API and we'd shelve the PR. Is there a look you rely on where achromatic→flat is intended?

**2. A standard control scrim/backer.** Controls floating over busy or low-contrast backdrops — and especially over a PlatformView where the capture shader can't run at all — need an opacity layer behind them to stay legible. We keep hand-rolling per-screen "backers" (a shape-matched scrim under the pill, light in light mode / dark in dark). Would you be open to a built-in scrim/backer on the glass controls (light/dark + strength)? It's the one piece that works everywhere including over PlatformViews, and we suspect it's something a lot of apps re-build.

No rush, but your read decides our wrap-up (and which staged PRs we send vs shelve). Thanks as always!

PR #106, opened 2026-06-17
Content-aware scroll-edge scrim (#102 fast-follow) + small GlassModalSheet / GlassEffect fixes

jfhair:
# Content-aware scroll-edge scrim (#102 fast-follow) + small GlassModalSheet / GlassEffect fixes

The fast-follow scrim lever from #103, plus a few small, independent fixes we'd been carrying app-side and wanted to upstream while we were in here. Kept as **separate commits** so they're reviewable (and splittable) individually:

1. `feat:` content-luminance scroll-edge scrim
2. `fix(GlassModalSheet):` handle-drag vs inner-scroll + `dragIndicatorColor`
3. `fix(GlassEffect):` defer backdrop capture when the boundary is mid-paint

If you'd rather take only some, the commits are independent — happy to split.

---

## 1. Content-luminance scroll-edge scrim (the #102 fast-follow)

The merged brightness machinery (#103) is the **discrete** lever — a WCAG contrast vote that flips a control's appearance *late*, only over genuinely dark content. This adds the **continuous** companion: the scroll-edge fade dissolves toward a color that tracks the content's luminance, so the edge stays seamless (and reduces edge contrast for floating controls) as content scrolls under the bars.

**What's added:**
- **`GlassContentAwareScope` exposes per-rect mean luminance.** `register()` gains `onLuminanceChanged` (and `onBrightnessChanged` becomes optional), delivered from the *same* single capture that serves the verdicts — N consumers, one readback. Deliveries are gated on >0.005 movement so consumer animations don't churn.
- **`GlassScrollEdgeEffect.contentAwareFade`.** Each enabled edge registers its own band with the scope and **lerps the fade's *target color* toward `darkFadeColor` by the band's content darkness** — fully dark at `luminanceDarkBelow` (0.45), off at `luminanceLightAbove` (0.72), animated over `fadeDuration` (280 ms). Over dark content the whole fade (including the strip above the bar) dissolves toward dark instead of toward the light page background; over light content it stays the light dissolve. Applied on **both** the captured-texture path (`GlassPage`) and the solid-color path. Bands are independent (top vs bottom), inert without a scope.
- **`GlassScaffold.contentAwareEdgeFade`** — one flag, composing with `contentAwareBrightness`.

> **Implementation note:** an earlier cut layered a separate dark gradient *over* the base fade, but the fade curve concentrates opacity at the bar edge (occluded by the bar), so the visible strip stayed light. Lerping the fade's target color instead darkens the whole band uniformly. Verified on-device (iOS, Impeller): over a black band the scrim dissolves to black (no light scrim); light↔dark cross-fades cleanly while the bar flips late.

**Also fixes a latent bug in shipped 0.16.0:** `GlassScaffold` wraps the body in the edge fade *first*, then in `GlassContentAwareContent` — so the fade overlays sit **inside** the captured region, and a `contentAwareBrightness` bar samples its own fade. This PR swaps the order (content-marker inside, fade outside) + a regression test. With the adaptive scrim that ordering is load-bearing (a scrim inside its own sampling input is a feedback loop).

The `content_aware_brightness_demo` turns the flag on so both levers show at once.

## 2. `fix(GlassModalSheet)` — two independent handle fixes

- **Handle drag no longer fights the inner scrollable.** A pointer-down in the top handle zone sets a `_handleDragActive` notifier (cleared on up/cancel), folded into the build's `AnimatedBuilder` listenable so it lands on pointer-down — before the inner `Scrollable`'s vertical-drag recognizer can claim the gesture. Gates a new `disableInnerScroll` on `_SheetLayout → _SheetContent`. Fixes a stall where, at the full state, grabbing the handle let the content overscroll and steal the pointer — acute over a PlatformView, which takes the touch natively mid-drag and freezes the sheet.
- **`dragIndicatorColor` now reaches the handle.** It was threaded into `_SheetLayout` but never passed to `_SheetHandleZone`/`_GlassDragIndicator` — a silent no-op. Wired through (`color ?? iOS-26 default`).

## 3. `fix(GlassEffect)` — defer capture when the boundary is mid-paint *(optional)*

`_captureBackgroundSync` runs `toImageSync` from `_handleTick` (transient phase, before paint), every frame during interaction. While the backdrop repaints, the boundary is dirty and `toImageSync` asserts in debug — caught, but it spams `[GlassEffect] toImageSync failed` nearly every interaction frame and drops that frame's capture. Guarded the same way `GlassScrollEdgeEffect` already does: if `boundary.debugNeedsPaint` (read inside an assert closure, so release never touches it), defer to a post-frame callback. ~20 lines, self-contained.

## Verification

- `flutter analyze` clean; full CI-equivalent suite **1926 tests** passing.
- Patch coverage **134/136**; the two uncovered lines are the GlassEffect deferred-capture branch — the headless test infra can't produce a mid-paint GPU capture (the same reason `lib/src/renderer/**` is excluded from the effective-coverage gate). Project effective coverage stays above the 90% gate. New/extended tests: the luminance output + adaptive fade (color + texture paths, early-darkening asserted directly), the wrap-order regression guard, and the modal-sheet handle-drag-cancel lifecycle.
- Golden failure set unchanged vs `main` (the known macOS-host set, CI-excluded).
- Scroll-edge scrim verified on-device; the GlassModalSheet fixes were validated in our app's map "park sheet" (the PlatformView stall they fix).

---

## Still open on our side (not blocking this PR)

Independent of all of the above, the two design questions in [this #103 comment](https://github.com/sdegenaar/liquid_glass_widgets/pull/103#issuecomment-4697884980) are still open — flagging again in case it got buried:

1. Now that `whitenStrength` exists, would you default achromatic tints to the luminosity (see-through) path — vs. the explicit `GlassTintBlend` enum we have staged?
2. Would you be open to a built-in control scrim/backer (light/dark + strength)?

No rush, and this PR doesn't depend on either.

### DIFF
--- a/example/lib/demos/content_aware_brightness_demo.dart
+++ b/example/lib/demos/content_aware_brightness_demo.dart
@@ -118,6 +118,10 @@ class _ContentAwareBrightnessDemoState
       topEdgeFade: true,
       bottomEdgeFade: true,
       contentAwareBrightness: true,
+      // The continuous companion: the scroll-edge fades darken early as
+      // medium/dark content scrolls under them (the App Store scrim), while
+      // the bar's contrast vote flips its appearance late.
+      contentAwareEdgeFade: true,
 
       // ── Body ──────────────────────────────────────────────────────────────
       body: CustomScrollView(
--- a/lib/widgets/overlays/shared/glass_modal_sheet_internal.dart
+++ b/lib/widgets/overlays/shared/glass_modal_sheet_internal.dart
@@ -27,6 +27,11 @@ class _SheetLayout extends StatelessWidget {
   final ScrollController scrollController;
   final ValueNotifier<SheetState> currentStateNotifier;
   final double expandProgressValue;
+
+  /// When true, the inner content scroll is force-disabled regardless
+  /// of expand state. Set during a handle drag so the scrollable can't
+  /// claim the vertical-drag gesture and fight the sheet drag.
+  final bool disableInnerScroll;
   final Widget child;
   final bool showDragIndicator;
   final Color? dragIndicatorColor;
@@ -67,6 +72,7 @@ class _SheetLayout extends StatelessWidget {
     required this.scrollController,
     required this.currentStateNotifier,
     required this.expandProgressValue,
+    this.disableInnerScroll = false,
     required this.child,
     required this.showDragIndicator,
     this.dragIndicatorColor,
@@ -83,11 +89,14 @@ class _SheetLayout extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final handleZone = _SheetHandleZone(indicatorWidth: dragIndicatorWidth);
+    final handleZone = _SheetHandleZone(
+      indicatorWidth: dragIndicatorWidth,
+      indicatorColor: dragIndicatorColor,
+    );
 
     final contentZone = _SheetContent(
       scrollController: scrollController,
-      isFullScreen: expandProgressValue > 0.95,
+      isFullScreen: expandProgressValue > 0.95 && !disableInnerScroll,
       padding: padding,
       child: child,
     );
@@ -332,9 +341,10 @@ class _SheetLayout extends StatelessWidget {
 // ===========================================================================
 
 class _SheetHandleZone extends StatelessWidget {
-  const _SheetHandleZone({required this.indicatorWidth});
+  const _SheetHandleZone({required this.indicatorWidth, this.indicatorColor});
 
   final double indicatorWidth;
+  final Color? indicatorColor;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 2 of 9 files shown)


sdegenaar:
Thanks again @jfhair,
great word and appreciated!!!

Good questions — quick takes:

1. Achromatic tints: prefer keeping it explicit for now — a GlassTintBlend.auto on the enum you have staged sounds right. Defaulting silently is too surprising for anyone with a deliberate near-white tint.

2. Control scrim/backer: interested, but before we add a new widget — does this fit as a scrimStrength param on GlassContentAwareBrightness? It already has the brightness verdict, same scope registration, feels like the natural home. If you need it to work outside the content-aware system then we can talk standalone widget, but want to avoid the surface area if we don't need it.

Open issues for both if you want to track them — happy to review whenever you have something staged.

Cheers

PR #112, opened 2026-06-19
fix(AdaptiveLiquidGlassLayer): preserve indicator morph across a platformViewBackdrop toggle

jfhair:
## Summary
Toggling `platformViewBackdrop` tears down the selected-tab indicator's morph — it snaps to the new tab instead of sliding.

## Root cause
`platformViewBackdrop` gates `useFullRenderer` in `AdaptiveLiquidGlassLayer`, which conditionally wraps the child in a `LiquidGlassBlendGroup`. Toggling the flag adds/removes that wrapper, changing the child's depth in the element tree. Because the child isn't behind a stable key, Flutter discards and re-inflates the whole child subtree, re-running `initState` on every animation controller inside it. In a bottom bar, the selected-indicator springs (`VelocitySpringBuilder`) are re-seeded at their already-settled value → no animation → snap.

(The indicator's `backgroundKey` swap, also driven by the flag, was a red herring — its setter only calls `markNeedsPaint()`.)

## Fix
Give the child a stable identity across the wrapper toggle: hold a `GlobalKey` in `State` and wrap the child in a `KeyedSubtree`. Flutter then reparents the subtree across the wrapper change instead of rebuilding it, so the live controllers survive and the morph plays. `AdaptiveLiquidGlassLayer` becomes a `StatefulWidget` to hold the key.

## Testing
Verified on a `GlassSearchableBottomBar` whose `platformViewBackdrop` toggles on tab change (over a Mapbox PlatformView vs Flutter content): before, the indicator snapped crossing that boundary; after, it morphs. Tabs that don't toggle the flag are unchanged.

### DIFF
--- a/lib/widgets/shared/adaptive_liquid_glass_layer.dart
+++ b/lib/widgets/shared/adaptive_liquid_glass_layer.dart
@@ -42,7 +42,7 @@ import 'inherited_liquid_glass.dart';
 ///   child: YourContent(), // Uses GlassTheme settings automatically
 /// )
 /// ```
-class AdaptiveLiquidGlassLayer extends StatelessWidget {
+class AdaptiveLiquidGlassLayer extends StatefulWidget {
   const AdaptiveLiquidGlassLayer({
     required this.child,
     this.shape = const LiquidRoundedSuperellipse(borderRadius: 0),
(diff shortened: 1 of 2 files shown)


jfhair:
Design direction for extending `platformViewBackdrop` to more widget types (`GlassContainer` / `GlassIconButton` / `GlassMenu`) is up for your steer here: https://github.com/sdegenaar/liquid_glass_widgets/discussions/113

Issue #115, opened 2026-06-21
[GlassBottomBar] please add new configurable fields for color of selected label and border radius of indicator


hnvn:
Currently, there's no way to change the color of selected label and border radius of pill indicator of GlassBottomBar. Please make those properties configurable. Thank you

sdegenaar:
Hey @hnvn,
Have added the properties. Try the latest v0.18.1. You can see the changes here :-
https://pub.dev/packages/liquid_glass_widgets/changelog

Thanks for raising and using the package. 
Let me know how you go.

Cheers

hnvn:
Thank you very much

PR #120, opened 2026-06-22
Add per-state label text style to bottom bars

jfhair:
## Summary

Adds `selectedLabelStyle` / `unselectedLabelStyle` (`TextStyle?`) to `GlassTabBar.bottom`, `GlassTabBar.searchable`, and the deprecated `GlassBottomBar` / `GlassSearchableBottomBar`.

## Why

0.18 added `selectedLabelColor` / `unselectedLabelColor` (from #115), which covers per-state label *color*. Matching Apple's tab bar also needs a per-state *font* — the selected label is heavier than a single shared `textStyle` can express (one `textStyle` applies to both states and overrides the color params). These two params let callers set the selected/unselected label's font family, weight, and letter-spacing per state.

## Behavior

The per-state style is **merged over** the base label style, so it overrides only the fields it sets and keeps the resolved per-state label color unless the style provides its own. Both default to `null` → existing behavior unchanged.

## Changes

Threads the two params public → layout → `BottomBarTabItem` (the shared label renderer for both bars), mirroring the existing `selectedLabelColor` path. The only render change is a `TextStyle.merge` at the label.

Related: #115.


hnvn:
I encountered this issue, if I provided both of `textStyle` and `selectedLabelColor`, the `selectedLabelColor` won't never be applied, the color from `textStyle` will override it.

jfhair:
Thanks @hnvn — that's a real one, and you've pinned it exactly.

The cause is in `BottomBarTabItem`'s label resolution (shipped in #115). The per-state colors only feed the **fallback** style:

```dart
final labelColor = selected ? (selectedLabelColor ?? selectedIconColor)
                            : (unselectedLabelColor ?? unselectedIconColor);
...
style: textStyle ?? TextStyle(color: labelColor, fontSize: labelFontSize, fontWeight: ...)
```

As soon as a `textStyle` is supplied, the `??` takes that branch and `labelColor` is dropped — so `selectedLabelColor` / `unselectedLabelColor` silently do nothing whenever `textStyle` is also set.

It's right next to this PR (#120 adds per-state `selectedLabelStyle` / `unselectedLabelStyle` in the same spot), so I'm happy to fold a fix in. The precedence I'd suggest is "more specific wins":

```dart
var base = textStyle ?? TextStyle(
  color: selected ? selectedIconColor : unselectedIconColor,
  fontSize: labelFontSize,
  fontWeight: selected ? FontWeight.w600 : FontWeight.w500,
);
final perStateColor = selected ? selectedLabelColor : unselectedLabelColor;
if (perStateColor != null) base = base.copyWith(color: perStateColor); // explicit color wins over textStyle's
final resolved = stateStyle != null ? base.merge(stateStyle) : base;   // per-state full style wins last
```

That makes an explicit `selectedLabelColor` apply even alongside `textStyle`, leaves `textStyle`-only callers unchanged, and composes with the `selectedLabelStyle` this PR introduces.

@sdegenaar — would you like me to roll this into #120 (≈3 lines + a test, in code this PR already touches), or keep #120 scoped to the new style param and tackle the `textStyle` / `selectedLabelColor` precedence yourself since it's a tweak to the #115 behavior? Happy either way.

sdegenaar:
Hey @jfhair,
thats will be fantastic and much appreciated. Roll it in if that makes it easier.

Thanks again. Really appreciating the contributions

Cheers

jfhair:
Done — rolled into #120.

`selectedLabelColor` / `unselectedLabelColor` now override the base label color even when a `textStyle` is supplied (previously they only fed the fallback style, so a `textStyle` silently dropped them). `textStyle`-only callers are unchanged, and it composes with the per-state `selectedLabelStyle` / `unselectedLabelStyle` this PR adds. Added a regression test asserting an explicit `selectedLabelColor` wins over a `textStyle` color.

Thanks @hnvn for the clean repro 🙏

Happy to help, @sdegenaar!

Issue #131, opened 2026-06-27
GlassTabBar.bottom: Incorrect Y-axis origin


l-7-l:
On Android, the frosted glass effect of GlassTabBar.bottom samples screen pixels from the top of the screen instead of the area directly beneath the bar. This seems to be an issue with coordinate starting points.

flutter doctor -v 
```
 flutter doctor -v
[✓] Flutter (Channel beta, 3.46.0-0.1.pre, on Arch Linux
    7.0.13-zen1-1-zen, locale zh_CN.UTF-8) [41ms]
    • Flutter version 3.46.0-0.1.pre on channel beta at
      /home/l-7-l/.tools/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision e5650835b0 (3 天前), 2026-06-23 22:00:24 -0700
    • Engine revision 300972dbc0
    • Dart version 3.13.0 (build 3.13.0-167.1.beta)
    • DevTools version 2.59.0
    • Feature flags: enable-web, enable-linux-desktop,
      enable-macos-desktop, enable-windows-desktop, enable-android,
      enable-ios, cli-animations, enable-native-assets,
      enable-swift-package-manager, omit-legacy-version-file,
      enable-lldb-debugging, enable-uiscene-migration

[✓] Android toolchain - develop for Android devices (Android SDK
    version 36.1.0) [3.5s]
    • Android SDK at /home/l-7-l/Android/Sdk
    • Emulator version 36.4.10.0 (build_id 15004761) (CL:N/A)
    • Platform android-36.1, build-tools 36.1.0
    • ANDROID_HOME = /home/l-7-l/Android/Sdk
    • Java binary at:
      /home/l-7-l/.tools/jetbrains/android-studio/jbr/bin/java
      This is the JDK bundled with the latest Android Studio
      installation on this machine.
      To manually set the JDK path, use: `flutter config
      --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build
      21.0.10+-14961533-b1163.108)
    • All Android licenses accepted.

[✓] Connected device (2 available) [280ms]
    • 21091116UC (mobile) • kfq8jfjvgipfnbbm • android-arm64 • Android
      13 (API 33)
    • Linux (desktop)     • linux            • linux-x64     • Arch
      Linux 7.0.13-zen1-1-zen

[✓] Network resources [1,227ms]
    • All expected network resources are available.

[...]

PR #133, opened 2026-06-27
Fix Android Impeller Y-axis origin

l-7-l:
#131


sdegenaar:
Hey @l-7-l . Thanks for the quick PR — you've correctly identified the root cause. However, we can't merge this yet for as 3.46 is still beta. Removing the flips now would break rendering for all users on the current stable channel (Flutter 3.44). We'll do this as a coordinated release once 3.46 hits stable.
Closing for now but tracking it — your analysis was spot on.

Cheers

PR #143, opened 2026-06-30
fix: support RTL in GlassTabBar.bottom indicator and hit-testing

### LINKED ISSUE #142: GlassTabBar.bottom: RTL not supported — pill and tap hit-testing land on the mirror-image tab
### Summary

`GlassTabBar.bottom` (and the deprecated `GlassBottomBar`) does not handle RTL `Directionality`. Under RTL the selected pill indicator and the tap/drag hit-testing land on the **mirror-image** tab.

### Why it happens

The indicator/gesture coordinate system and `AnimatedGlassIndicator` operate in **physical, left-anchored alignment space** (`x == -1` is always the left edge):

- `TabDragGestureMixin.computeTabAlignment(index)` → `DraggableIndicatorPhysics.computeAlignment(index, count)` maps index `0` to the physical **left**.
- `onBarTapDown` / `onBarDragEnd` map a physical pointer position back to an index assuming the same left-anchored space.

But the tab `Row`s in `tab_bar_bottom_layout.dart` use a plain `Row`, which honours the ambient `Directionality` and visually **reverses** under RTL. So the first tab is drawn on the right, while the indicator (and hit-testing) still treat index `0` as the left edge — they disagree.

### Steps to reproduce

```dart
Directionality(
  textDirection: TextDirection.rtl,
  child: GlassTabBar.bottom(
    tabs: const [
      GlassTab(label: 'Home',    icon: Icon(CupertinoIcons.home)),
      GlassTab(label: 'Search',  icon: Icon(CupertinoIcons.search)),
      GlassTab(label: 'Profile', icon: Icon(CupertinoIcons.person)),
    ],
    selectedIndex: 0,
    onTabSelected: (i) => print(i),
  ),
)
```

### Expected

- First tab (`Home`) renders on the trailing (right) edge.
- The pill sits on the selected tab.
- Tapping `Home` reports index `0`.

### Actual

- `Home` renders on the right, but the pill is drawn on the **left** (index 0 → physical left).
- Tapping `Home` (right edge) reports the **last** index (`2`) instead of `0`.

### Workaround

Consumers currently have to force `Directionality.ltr` and reverse the tab list + index + callback by hand, e.g.:

```dart
Directionality(
  textDirection: TextDirection.ltr,
  child: GlassTabBar.bottom(
    tabs: tabs.reversed.toList(),
    selectedIndex: count - 1 - currentIndex,
    onTabSelected: (i) => onTap(count - 1 - i),
    ...
  ),
)
```

This should be handled inside the widget.

### Environment

- liquid_glass_widgets: 0.19.5

I have a fix + regression test ready and will open a PR referencing this issue.

naeemeltaief:
## What

Fixes RTL support for `GlassTabBar.bottom` (and the deprecated `GlassBottomBar` shim).

Closes #142.

## Problem

The indicator/gesture coordinate system and `AnimatedGlassIndicator` operate in **physical, left-anchored** alignment space (`x == -1` is always the left edge):

- `computeTabAlignment(index)` → `DraggableIndicatorPhysics.computeAlignment(index, count)` maps index `0` to the physical left.
- `onBarTapDown` / `onBarDragEnd` map a physical pointer position back to an index in the same space.

But the tab `Row`s in `tab_bar_bottom_layout.dart` use a plain `Row`, which honours the ambient `Directionality` and visually **reverses** under RTL. The two disagree, so under RTL the selected pill and the tap/drag hit-testing land on the **mirror-image** tab — tapping the first (right-edge) tab selects the last index.

This forced consumers to wrap the bar in `Directionality.ltr` and reverse the tab list, selected index and callback by hand.

## Fix

Contained to `tab_bar_bottom_layout.dart`. In `_buildBar`, when the ambient direction is RTL:

- reverse the tab data,
- mirror the selected index (`count - 1 - selectedIndex`),
- mirror the tap callback (`onTabSelected(count - 1 - i)`),
- pin the bar's internals to `Directionality.ltr` so the physical coordinate space stays consistent with the (now-normalised) tab order.

Net effect under RTL: correct ordering — the first tab sits on the trailing (right) edge — with the pill and hit-testing aligned to it. LTR is an exact no-op (same widget tree as before). All physics (jelly drag, spring, sway) keep running in their original LTR space, so behaviour is unchanged there.

## Tests

Adds `test/widgets/surfaces/tab_bar_bottom_layout_rtl_test.dart`:

- **tap → logical index** under RTL: tapping `Home` (rendered on the right) reports index `0`; tapping `Profile` (left) reports `2`. This test **fails on `main`** (reports the mirrored index) and passes with the fix.
- **ordering**: the first tab renders to the right of the last tab and past the horizontal centre.

Existing bottom-bar suites (`glass_bottom_bar_test.dart`, `glass_bottom_bar_drag_test.dart`, `glass_bottom_bar_coverage_test.dart`, `glass_tab_bar_constructors_test.dart`) all still pass. `flutter analyze` is clean.

## Notes

- Scoped to the bottom placement (the reported case). The searchable layout likely needs the same treatment; happy to follow up in a separate PR if you'd like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+# Unreleased
+
+- **`GlassTabBar.bottom` / `GlassBottomBar`** — fixes RTL support (#142, @naeemeltaief).
+  The indicator/gesture coordinate system operates in physical (left-anchored) alignment
+  space while the tab `Row`s honour the ambient `Directionality`, so under RTL the pill and
+  the tap/drag hit-testing landed on the mirror-image tab. The bottom layout now normalises
+  the tab order, selected index and tap callback for RTL and pins *only the tab rows* to LTR,
+  so the first tab sits on the trailing edge with the pill and hit-testing aligned to it.
+  `indicatorExpansion` / `tabPadding` continue to resolve against the ambient direction, so
+  `EdgeInsetsDirectional` values still work in RTL. Consumers no longer need to force
+  `Directionality.ltr` and reverse tabs by hand.
+
 # 0.19.5
 
 - **`LiquidGlassSettings`** — adds `platformViewFallbackColor` (#138, @jfhair). Splits
--- a/lib/widgets/surfaces/shared/tab_bar_bottom_layout.dart
+++ b/lib/widgets/surfaces/shared/tab_bar_bottom_layout.dart
@@ -144,6 +144,19 @@ class _TabBarBottomLayoutState extends State<TabBarBottomLayout> {
   // Both bars reference kBottomBarGlassDefaults so their glass is guaranteed identical.
   static const _defaultGlassSettings = kBottomBarGlassDefaults;
 
+  /// Lays out the tab [Row] in physical (LTR) order regardless of the ambient
+  /// direction, so the first child is on the left — matching the indicator and
+  /// gesture coordinate space. RTL ordering is carried by the reversed tab data
+  /// in [_buildBar], not by the ambient direction of these Rows. Scoping the
+  /// pin to the Rows keeps `Directionality.of(context)` intact for the rest of
+  /// the subtree (notably [indicatorExpansion] / [tabPadding] resolution).
+  static Widget _ltrTabRow({required List<Widget> children}) {
+    return Directionality(
+      textDirection: TextDirection.ltr,
+      child: Row(children: children),
+    );
+  }
+
   @override
   Widget build(BuildContext context) {
     if (!widget.adaptiveBrightness && widget.brightnessOverride == null) {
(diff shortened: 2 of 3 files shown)


sdegenaar:
Hey @naeemeltaief,
Good work on this — the bug is real and the normalisation approach is the right way to handle it.

One thing to verify before we merge: the PR description mentions pinning the entire bar to Directionality.ltr. If that's the case it's too broad — indicatorExpansion and tabPadding are both resolved against Directionality.of(context) inside TabIndicator, so wrapping the full bar will break any RTL consumer passing EdgeInsetsDirectional values. The pin should only be on the two inner Row widgets, not the whole bar subtree. Can you confirm the actual scope of the wrap?

Also agreeing with Copilot on swapping .first for .hitTestable() in the tests — minor but the right call given the overlapping stack. Happy to review again with changes and confirmation.

Cheers

naeemeltaief:
Thanks for the careful review — both points were spot on.

**Scope of the LTR pin.** You're right, wrapping the whole bar was too broad. `TabIndicator` resolves `indicatorExpansion` and `tabPadding` against `Directionality.of(context)`, so the full-subtree wrap would have forced those to LTR and silently dropped any `EdgeInsetsDirectional` an RTL consumer passes.

I've narrowed the pin to just the two tab `Row`s via a small `_ltrTabRow` helper. The reasoning that lets this work: the indicator alignment (`computeAlignment(index, count)`), the `AnimatedGlassIndicator` position, and the gesture math (`alignmentFromGlobal` off the render box) are all already direction-independent / physical — the *only* direction-sensitive piece was the tab Rows reversing under RTL. So pinning just the Rows is enough to line their physical order up with the coordinate space, and `Directionality.of(context)` stays intact for the rest of the subtree. Confirmed `indicatorExpansion` / `tabPadding` now resolve against the ambient direction again.

**Tests.** Switched to `.hitTestable()`. One wrinkle worth flagging: with `MaskingQuality.off` every label is drawn twice — the base unselected row plus the vibrant selected-overlay row — so each label still matches two hit-testable widgets and a bare `.hitTestable()` is ambiguous for `tap()`/`getCenter()`. I settled on `.hitTestable().first`, which keeps the occlusion-safe filtering while deterministically resolving to the base row (whose text sits at the tab's real on-screen position). Open to a different approach if you'd prefer.

I also re-verified the regression test against pristine `0.19.5`: the tap-to-index test fails there (tapping the first tab reports the last index) and passes with the fix. Full bottom-bar suite (119 tests) stays green and `flutter analyze` is clean. CHANGELOG updated to reflect the narrower scope.

Branch is force-pushed with the amended commit. Thanks again!

sdegenaar:
Hey @naeemeltaief,
Thank you so much for your contribution here. It has been committed as part of v0.19.7 and you have been credited in the changelog. Thanks again for your great work. These contribution really help!

Cheers

PR #162, opened 2026-07-16
feat(effects): add ProgressiveBlur — graduated backdrop-blur primitive

Ahmadre:
## Important | Merge order

Please merge in this order:
1. #161  -> bumps patch to: 0.21.7
2. #163  -> bumps patch to: 0.21.8
3. (this PR) -> bumps minor to: 0.22.0

> 0.21.7 -> 0.21.8 -> 0.28.0

## Summary / New Component

IMG_4075

Adds **`ProgressiveBlur`** — a *graduated* (progressive) backdrop blur: a clean
gaussian frost that is strongest at one edge and eases to perfectly sharp at the
opposite edge. It's the Signal / iOS-26 header dissolve — stack it behind a
translucent app bar or bottom bar so scrolling content dissolves beneath the bar
instead of ending on a hard cut-off.

Today the library's glass surfaces apply a **uniform** blur. This is the one
progressive primitive it's missing. `ProgressiveBlur` is self-contained — it
needs no `LiquidGlassLayer` or glass ancestor — so it can back any bar.

```dart
Stack(
  children: [
    const Positioned(
      top: 0, left: 0, right: 0, height: 96,
      child: ProgressiveBlur(maxSigma: 20),
    ),
    // ... your translucent app bar on top ...
  ],
)
```

## Related Issue:

- #160 

## Public API

| Symbol | Notes |
| --- | --- |
| `ProgressiveBlur` | `StatefulWidget`; `maxSigma` (default `18`), `direction` (default `topToBottom`), `falloff` (default `1.2`). |
| `ProgressiveBlurDirection` | `topToBottom` / `bottomToTop` / `leftToRight` / `rightToLeft` — the edge the blur is strongest at. |
| `ProgressiveBlur.preload()` | Static; pre-compiles the shader. Idempotent, never throws. Optional — `initialize()` already warms it. |

`maxSigma: 0` is a passthrough (no blur), so you can drive it straight off a
scroll offset to fade the frost in and out.

## Why the obvious recipe doesn't work (and what this does instead)

The intuitive approach — "blur the backdrop, then fade it out with a
`ShaderMask` / gradient mask" — **fails on Impeller**: a `BackdropFilter`'s
captured backdrop is not part of an ancestor `ShaderMask`'s layer, so the mask
has nothing to reveal and the bar shows no blur at all.

Instead this binds a fragment shader **as the `ImageFilter` of the
`BackdropFilter` itself** (`ui.ImageFilter.shader`), so the engine hands the
[...]

PR #163, opened 2026-07-16
fix(overlays): correct GlassPopover positioning in nested overlays

Ahmadre:
## Important | Merge order

Please merge in this order:
1. #161  -> bumps patch to: 0.21.7
2. (This PR) -> bumps patch to: 0.21.8
3. #162 -> bumps minor to: 0.22.0

> 0.21.7 -> 0.21.8 -> 0.28.0

## Summary / Fix

Two `GlassPopover` positioning bugs, both in how the morphing overlay is placed:

1. **The popover drifts off its trigger when there's a nested overlay** between
   it and the screen.
5. **An intrinsic-height popover overflows when its content grows while open.**

Both are backwards-compatible fixes — no API change. Regression tests included
for each.

## 1 · Nested-overlay offset → render into the **root** overlay

The morphing overlay is positioned with **absolute, root-relative coordinates**
— it reads the trigger's `localToGlobal(Offset.zero)` and places the morph at
that global point. But the `OverlayPortal` attached to the **nearest** overlay
(`OverlayChildLocation.nearestOverlay`, the default).

When the nearest overlay is itself offset from the screen origin — e.g. a
`ShellRoute` / nested-`Navigator` content area sitting beside a persistent side
rail — that overlay's origin gets counted **twice**: once by the global trigger
coordinate, once by the overlay's own offset. The popover lands shifted away
from the button that opened it (by the width of the rail, in the classic case).

**Fix:** target the root overlay, which always coincides with the global
coordinate space the morph is computed in:

```dart
OverlayPortal(
  controller: _overlayController,
  overlayLocation: OverlayChildLocation.rootOverlay, // was: nearest (default)
  overlayChildBuilder: _buildMorphingOverlay,
)
```

Top-level usage (no nested overlay) is unaffected — there the nearest overlay
*is* the root overlay.

## 2 · Intrinsic-height popover froze its content height

With no `popoverHeight`, the popover measures its content once at open time and
uses that as `maxHeight`. Content that grows **after** open — a row expanding
after a toggle, an async list filling in, a validation message appearing —
overflows the frozen box.

**Fix:** wrap the intrinsic-mode content in a `SizeChangedLayoutNotifier` and
re-measure on the next frame when it reports a size change, so the popover
follows the live content height (bounded by the screen) instead of clamping to
the open-time measurement:

```dart
NotificationListener<SizeChangedLayoutNotification>(
  onNotification: (_) { _scheduleRemeasure(); return true; },
  child: SizeChangedLayoutNotifier(child: SizedBox(key: _contentKey, ...)),
)
```

`_scheduleRemeasure()` coalesces to one post-frame callback, re-reads the
content `RenderBox`, and `setState`s the new height only when it actually
changed. The intrinsic `maxHeight` is now the screen-bounded
`_getMaxPopoverHeight()` (already in the file) rather than the frozen
measurement, so the content can re-layout to its new natural height for the
notifier to report.

Fixed-`popoverHeight` popovers are untouched — the caller owns scrolling there.

## Tests

`test/widgets/overlays/glass_popover_positioning_test.dart`:

- **root-overlay wiring** — asserts the popover's `OverlayPortal.overlayLocation`
  is `OverlayChildLocation.rootOverlay` (deterministic, no animation).
- **live re-measure** — opens an intrinsic-height popover whose content is 80 px,
  grows it to 260 px while open, and asserts no overflow and that the popover
  followed to 260 px (fails on the pre-fix frozen-height behaviour).

Both pass; the existing `glass_popover_test.dart` suite stays green.

## Versioning

Patch bump → **`0.21.8`**. Pure bug fixes, no API surface change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,29 @@
+# 0.21.8
+
+## 🐛 Bug Fixes
+
+- **`GlassPopover` drifted off its trigger inside a nested overlay** — the
+  morphing overlay is positioned with absolute, root-relative coordinates (the
+  trigger's `localToGlobal(Offset.zero)`), but its `OverlayPortal` attached to
+  the *nearest* overlay. When that nearest overlay is itself offset from the
+  screen origin — e.g. a `ShellRoute` / nested-`Navigator` content area sitting
+  beside a side rail — its origin is counted twice and the popover appears
+  shifted away from the button that opened it. The portal now targets
+  `OverlayChildLocation.rootOverlay`, which always shares the global coordinate
+  space, so the popover lands on its trigger in every embedding. Top-level usage
+  (no nested overlay) is unaffected.
+
+- **Intrinsic-height `GlassPopover` overflowed when its content grew while open**
+  — with no `popoverHeight`, the popover measured its content once at open time
+  and froze that height. Content that grew afterwards (a row expanding after a
+  toggle, an async list filling in) overflowed the frozen box. The popover now
+  wraps its content in a `SizeChangedLayoutNotifier` and re-measures on the next
+  frame, so it follows the live content height (bounded by the screen) instead of
+  clamping to the open-time measurement. Fixed-`popoverHeight` popovers are
+  unchanged (the caller owns scrolling there).
+
+---
+
 # 0.21.6
 
 ## 🐛 Bug Fixes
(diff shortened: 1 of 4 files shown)


sdegenaar:
Hey @Ahmadre. Thank you so much!!!

Integrated via cherry-pick onto main in fe27274 (0.21.8) due to the fork not having "Allow maintainer edits" enabled. Changes are fully included — closing manually.

Cheers

Issue #136, opened 2026-06-29
Regression: GlassSegmentedControl selected pill invisible inside a GlassContainer (0.18+)


jfhair:
**Summary**

A `GlassSegmentedControl`'s selected-pill fill stopped rendering when the control sits inside a `GlassContainer` (or any `avoidsRefraction` scope) — the selection is invisible at rest. This worked on ≤0.17 and regressed in 0.18+. The class docs explicitly list placing it inside glass containers as a supported pattern.

**Root cause (best I can tell)** — two independently-added pieces collide:

- `GlassContainer` wraps its child in `InheritedLiquidGlass(avoidsRefraction: true)` ("Containers block children from refracting background") — since `1e58e4c` (2026-06-08).
- `AnimatedGlassIndicator`'s resting background pill now returns `SizedBox.expand()` whenever `avoidsRefraction` is true — added in `3aedde39` (2026-06-22), to keep the pill out of a `GlassEffect` capture snapshot.

The indicator can't distinguish a *transient capture* from a container's *permanent* `avoidsRefraction`, so inside a `GlassContainer` the pill is hidden for good. Since the glass pass is gated on `fade > 0.05` (drag/press only), the resting fill is the *only* thing showing the selection — so it vanishes entirely at rest.

**Repro:** `GlassSegmentedControl` (premium) inside `GlassContainer(useOwnLayer: true)`, any `indicatorColor` — pill doesn't render (confirmed even with fully-opaque red). Remove the container → it renders.

**Preferred direction?**

1. Is the `3aedde39` pill-hide meant to be **capture-only**? If there's (or could cheaply be) a capture-specific signal distinct from a container's static `avoidsRefraction`, gating the hide on that would fix it generally and restore the documented "inside a glass container" pattern.
2. Or should **`useOwnLayer` render its own track-surface glass**? Today it only groups a layer (`AdaptiveLiquidGlassLayer` paints no surface), so Standalone Mode gives a glass pill but no glass track — meaning glass-on-the-track still needs a container wrapper.

Happy to send a PR for whichever you prefer.

sdegenaar:
Hey @jfhair — thanks for the detailed diagnosis, you are correct on the regression.

[...]

Issue #77, opened 2026-06-02
Liquid Glass in Glass Segment Control


justbetterApps:
Hi there, i found out that the liquid or blur effects in GlassSegmentControl are not working, while i am dragging the indicator the liquid effect appears, other than that there is no blur or liquid effects in the whole widget

sdegenaar:
Hey @justbetterApps,
By default this control is standard quality as usually it's in a scrollable list. Have you tried setting to premium? If in a scrollable region, also set useOwnLayer: true:

```dart
GlassSegmentedControl(
  quality: GlassQuality.premium,
  useOwnLayer: true,
  segments: [...],
  ...
)
```

Also make sure your widgets are inside a GlassPage with a background (gradient, image, etc.) — the glass effects refract and blur against whatever is behind them, so without a rich background they can appear flat or invisible, even at standard quality.

Also, the upcoming release includes GlassScaffold which handles all the background/layer setup automatically — that should make this much simpler.

Cheers

sdegenaar:
Hey @justbetterApps 
How did you go with this. Did premium quality solve it for you. If not let me know and if you have screens/images what you wan to achieve.

Cheers

justbetterApps:
<img width="1080" height="450" alt="Image" src="https://github.com/user-attachments/assets/7cb16dab-6871-49a9-b8b2-1b1343c1840d" />
<img width="1078" height="482" alt="Image" src="https://github.com/user-attachments/assets/e3d12099-dcec-47a4-bb4f-304f36cb4e6f" />

Hey @sdegenaar  the glass sliding segment and glass tab bar both have this issue, i dont know whats the issue but even with premium quality as you can already see the liquid or blur effects are not there, however when i am pressing the thumb on the indicator there are glass and blur effects, so right now i am using glassbottom bar as a tabbar as it works as expected

sdegenaar:
Hi @justbetterApps,
It does need a little more polish in the glass interactive pill but you are talking about the segment/tab bar underneath is that right. From what I can see the actual ios26 Segmented is just an opaque background like we have. 

<img width="344" height="285" alt="Image" src="https://github.com/user-attachments/assets/8dc2be6a-ec46-403c-aaa6-808d6d54b9b7" />

This is running through a platform view to comapre with a background so we can see. 

<img width="472" height="389" alt="Image" src="https://github.com/user-attachments/assets/578e465d-3995-44ab-b4ae-975d2f9d9ac8" />

[...]

PR #164, opened 2026-07-19
Fix GlassSegmentedControl track glass regression

F1orian:
## Summary

This fixes the fixed-size `GlassSegmentedControl` track so it renders as actual glass instead of a plain decorated box.

Changes:

- render the segmented-control track through package-owned `AdaptiveGlass`
- resolve track `settings` and `quality` through `GlassThemeHelpers`, so inherited/global/theme settings apply unless explicitly overridden
- preserve `backgroundColor` precedence by mapping it to the track backer color
- keep the animated/moving indicator as a sibling above the track so its jelly expansion is not clipped by the track shape
- add regression tests for the default glass track, inherited minimal/theme settings, and indicator layering

## Root cause

`GlassSegmentedControl` had glass configuration parameters, but the fixed-size track path only used a `DecoratedBox`/color backer for the outer track. As a result, app-level glass settings such as blur/quality could affect the indicator while the track background itself stayed non-glass.

## Validation

- `flutter test test/widgets/interactive/glass_segmented_control_test.dart test/widgets/interactive/glass_segmented_control_vertical_test.dart`
- `flutter analyze`

### DIFF
--- a/lib/widgets/interactive/glass_segmented_control.dart
+++ b/lib/widgets/interactive/glass_segmented_control.dart
@@ -417,7 +417,8 @@ class GlassSegmentedControl extends StatefulWidget {
 
   /// Glass effect settings for the segmented-control track.
   ///
-  /// If null, uses optimized defaults:
+  /// If null, resolves from inherited/global/theme settings before falling
+  /// back to optimized defaults:
   /// - thickness: 30
   /// - blur: 3
   /// - chromaticAberration: 0.5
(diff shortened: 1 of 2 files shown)


sdegenaar:
Hey @F1orian,
Thanks for all your work on this. We did a bit if research on this back and forth here https://github.com/sdegenaar/liquid_glass_widgets/issues/136 and
https://github.com/sdegenaar/liquid_glass_widgets/issues/77. We concluded ios26 does not have glass in the track on UISegmentedControl and we did GlassTabBar.inline if you wanted it like that. Have a read and let me know if that addresses the issue for you. 

Cheers

F1orian:
Hi @sdegenaar, 
thanks for taking the time to review this. I had a look at the implemented solution and it now works perfectly for our usecase. 

Cheers!

PR #176, opened 2026-07-27
Fix bottom bar collapse trajectory into the extra button area

jingluoguo:
## Summary

This PR fixes the collapse trajectory of `GlassBottomBar` when collapsing towards the extra button.

Changes included:
- align the collapsed tab pill with the extra button when using `GlassBottomBarCollapseDirection.towardsExtraButton`
- remove the visible gap between the collapsing pill and the extra button
- adjust stacking order so the extra button overlays correctly during the collapse animation
- add regression tests covering collapse direction, left/right extra button placement, scroll-driven collapse/expand, tap-to-expand behavior, and extra button scaling

## Why

When the bottom bar collapsed towards the extra button, the tab pill could enter the extra area with an incorrect trajectory and leave an unintended visual gap. This made the animation feel off, especially when the extra button was used as the collapse target.

## Testing

- Added widget tests for collapse positioning and animation behavior
- Added regression coverage for:
  - collapse towards extra button
  - left extra button placement
  - scroll up collapse / scroll down expand
  - tap to expand before firing extra action
  - extra button scale during collapse


sdegenaar:
Hey @jingluoguo — great fix, well tested, and cleanly done. 🙌 Merged into 0.24.4 and we added a dedicated demo to the example app so everyone can see it in action. Your name's in the changelog — thanks for the contribution!

Cheers

PR #179, opened 2026-07-28
Fix: GlassMenu mispositions inside nested Navigators

sinanhaci:
## Problem

`GlassMenu` positions its popup incorrectly when it lives inside a nested `Navigator` whose `Overlay` isn't flush against the screen's true top-left corner — e.g. a persistent side rail/drawer next to a `StatefulShellRoute.indexedStack` branch, where each branch owns its own `Navigator` (and therefore its own `Overlay`) offset from the screen edge by the rail's width.

### Root cause

`_openMenu()` (`lib/widgets/overlays/shared/glass_menu_internal.dart`) captures the trigger's anchor with:

```dart
_triggerGlobalPosition = renderBox.localToGlobal(Offset.zero); // no `ancestor` arg
```

Called with no `ancestor`, `localToGlobal` resolves to **screen-global** coordinates (relative to the root `RenderView`).

The overlay content built in `_buildMorphingOverlay` then positions itself with `Positioned(left: blobBLeft, top: blobBTop, ...)` inside whichever `Overlay` hosts it — and `OverlayPortal` (used with no explicit `overlayLocation`) defaults to `OverlayChildLocation.nearestOverlay`.

These two coordinate spaces only agree when the nearest `Overlay`'s own origin coincides with the screen's true `(0, 0)`. That's true for a plain single-`Navigator` app, but breaks as soon as `GlassMenu` is used inside a *nested* `Navigator` positioned away from the screen edge: the popup renders shifted by exactly that `Overlay`'s offset from the screen, and the drift grows with however far that offset is (in our repro, a collapsible side rail — the wider it opens, the further the menu drifts from its trigger).

### Fix

Target `OverlayChildLocation.rootOverlay` explicitly in the `OverlayPortal` call, so the popup's coordinate space always matches the anchor's regardless of how deeply nested the trigger's own `Navigator` is.

### Testing

- Existing package test suite (`glass_menu_test.dart`, `glass_menu_primitives_test.dart`, `glass_menu_item_coverage_test.dart` — 69 tests) passes unchanged.
- Verified the fix in a real app with a `StatefulShellRoute.indexedStack` shell (persistent collapsible side rail, 80px collapsed / 280px expanded) where `GlassMenu` triggers inside a branch's list items were drifting proportionally to the rail's width; after the fix the popup anchors correctly at both rail states.

Added a changelog entry describing the fix under `# Unreleased`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,21 @@
+# Unreleased
+
+## 🐛 Bug Fixes
+
+- **`GlassMenu` mispositioned inside nested `Navigator`s** — the menu's popup
+  is now hosted on the root `Overlay` instead of the nearest one.
+  `_openMenu()` captures the trigger's anchor via
+  `renderBox.localToGlobal(Offset.zero)`, which resolves to screen-global
+  coordinates, but the overlay content built in `_buildMorphingOverlay` was
+  positioned inside whichever `Overlay` happened to be nearest — the default
+  target for `OverlayPortal`. Those only agree when the nearest `Overlay`
+  sits at the screen's true origin. Any layout with its own nested
+  `Navigator`s not flush against the screen edge (e.g. a persistent side
+  rail/drawer next to a `StatefulShellRoute.indexedStack` branch, each branch
+  owning its own `Navigator`/`Overlay`) broke this assumption: the menu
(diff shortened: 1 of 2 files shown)


sdegenaar:
Hey @sinanhaci. Thank you for taking the time to contribute. Very much appreciated! Release in v0.25.1

Cheers

PR #180, opened 2026-07-30
feat: prevent adaptive glass quality oscillation

jingluoguo:
## Summary
This PR adds an opt-in way to stabilize adaptive glass quality when temporary performance fluctuations cause repeated transitions between liquid glass and frosted glass.
Changes included:
add allowStepUp control for adaptive quality recovery
allow quality to degrade while preventing upgrades when allowStepUp is false
preserve the downgraded quality after a reset warm-up
keep the default behavior unchanged with allowStepUp: true
improve API documentation for the new behavior
add regression coverage for locked quality after warm-up
## Why
When device performance temporarily drops, the adaptive quality system can downgrade from liquid glass to frosted glass and then upgrade again shortly afterward. This repeated visual switching feels unstable and makes the interface look inconsistent.
allowStepUp: false lets applications prefer a stable lower quality after degradation, while still allowing the system to downgrade when performance requires it.
## Testing
Added a regression test covering:
quality degradation with allowStepUp: false
reset-triggered warm-up attempting to promote quality
preserving the previously downgraded quality
Also verified that flutter analyze passes for the affected files. The full flutter test suite has not been run yet.

### DIFF
--- a/lib/utils/glass_quality_adapter.dart
+++ b/lib/utils/glass_quality_adapter.dart
@@ -130,10 +130,11 @@ class GlassQualityAdapter {
   /// from this value.
   final int targetFrameMs;
 
-  /// When `true`, the adapter may step quality **up** after a sustained period
-  /// of good performance. When `false` (the default), only step-downs occur —
-  /// the starting quality from Phase 2 is the permanent ceiling for the
-  /// session (or until [reset] is called).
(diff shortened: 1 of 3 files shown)


sdegenaar:
Hey @jingluoguo
Thanks for the finding and fixing this. Much appreciated. Will be part of v0.26.1 releasing shortly.

Cheers

PR #197, opened 2026-08-09
fix(sheet): stale _currentState, overscroll axis guard, one-shot axis lock

### LINKED ISSUE #196: GlassModalSheet: sheet refuses to expand after an aborted drag (`_currentState` left holding a mid-drag prediction)
## 1. Sheet permanently refuses to expand after an aborted drag  *(bug — has a fix)*

**Symptom.** After a drag that starts upward and settles back where it began,
the sheet stops responding to upward drags entirely. It still responds to
downward drags, and one downward drag restores normal behaviour. To the user
the sheet looks frozen at `half`.

**Cause.** `_currentState` is written from three places, and only one of them
means "settled":

- `_snapToState` (`glass_modal_sheet_state.dart:229`) — the real transition.
- `_jumpTo` (:256) and `_applyDrag` (:448) — set `_currentState = target`
  mid-drag, where `target` comes from `resolveTarget(...)`. That is a
  *prediction* of where the drag would land, not where the sheet is.

`_snapToState` reconciles `_currentState` only inside `if (state != _settledState)`:

```dart
if (state != _settledState) {
  ...haptics...
  _currentState = state;
  _settledState = state;
  widget.onStateChanged?.call(state);
}
```

So when a drag predicts `full`, then settles back to `half` — the state it
started from — the guard is false, the body is skipped, and `_currentState`
keeps the stale `full`. `_settledState` and the rendered position both say
`half`.

`evaluateMove` then takes its topmost-detent branch:

```dart
if (currentState == maxState) {      // full == full
  if ((y - dragStartY) < 0) {        // swiping up
    if (hasScrollClients) {
      phase = GesturePhase.scrolling;
      return false;                  // refused
```

Downward drags fall to the `else` branch, get claimed, and `_applyDrag`
re-predicts `half`, which silently resyncs `_currentState`. Hence
"drag down once and it works again."

`onStateChanged` never fires for any of this, so a host app mirroring sheet
state cannot detect or work around it.

**Instrumented capture** (probe inside `_onPointerMove`, sheet resting at
`half`, user dragging up 169pt):

```
p=scrolling  full/full  dy=-169  dx=3  sc=1  claim=0
```

`currentState/maxState` both read `full` while the sheet is rendered at `half`.

**Reproduce.** Sheet with `detents: {medium, large}` at `half`:
1. Press on the sheet and drag up ~30–60pt — past the 10pt threshold, short
   of the snap threshold.
2. Release. The sheet settles back to `half`.
3. Try to drag up again. Refused, indefinitely. A downward drag clears it.

**Fix** (branch `fix/sheet-stale-current-state`, one line moved): reconcile
`_currentState` on every snap, leaving the side effects gated on
`_settledState` as intended.

```diff
+    // Reconcile on EVERY snap: _applyDrag/_jumpTo mutate _currentState to the
+    // in-flight predicted target, so a drag ending at the state it started
+    // from leaves that prediction behind.
+    _currentState = state;
+
     if (state != _settledState) {
       ...haptics...
-      _currentState = state;
       _settledState = state;
       widget.onStateChanged?.call(state);
```

All 148 existing sheet tests pass.

---

## 2. `_onScrollNotification` ignores the notification's axis  *(bug)*

`glass_modal_sheet_state.dart` reacts to any `OverscrollNotification` with
`overscroll < 0` by claiming the gesture and re-anchoring the vertical drag
origin to the finger's current y:

```dart
if (notification is OverscrollNotification && notification.overscroll < 0) {
  _gestureArena.phase = GesturePhase.contentDrag;
  _gestureArena.dragStartY = notification.dragDetails!.globalPosition.dy;
```

Scroll notifications bubble from **any** descendant, including horizontal
ones. A horizontal list inside the sheet pulled past its leading edge emits
exactly this, so scrolling sideways drags the sheet vertically, and the
re-anchored origin can keep `dy` below the 10pt threshold.

`WeatherCard`-style widgets can guard themselves by consuming horizontal
notifications, but every host has to know to do it for every horizontal
scrollable it mounts.

**Suggested fix:** ignore notifications whose `metrics.axis != Axis.vertical`.

---

## 3. Axis test is cumulative from the touch origin  *(design question)*

`evaluateMove` compares total travel since pointer-down:

```dart
final dy = (y - dragStartY).abs();
final dx = (x - dragStartX).abs();
if (dy > threshold && dy > dx) {
```

Two consequences:

- **Long horizontal scrolls suppress the sheet for the rest of the touch.**
  After ~220pt sideways on a 440pt-wide screen, the sheet needs >220pt of
  vertical travel before it will respond.
- **`phase = contentDrag` latches and is never re-evaluated.** A sideways
  swipe that begins with a slight vertical wobble drags the sheet for its
  whole duration.

Net effect is that a horizontal scroll sometimes moves the sheet and
sometimes doesn't, depending on the first ~10pt. Would you consider a
one-shot axis decision (lock the axis once, per touch) rather than a
cumulative comparison? Happy to open a PR if you'd like it changed — filing
as a question since it's a behaviour choice, not clearly a defect.

---

## Notes

- Reproduced on iOS, physical device, `liquid_glass_widgets` 0.29.2.
- #1 and #2 both produce "the sheet won't expand", but they are independent:
  #1 was captured with zero horizontal overscrolls in play.

jfhair:
Fixes #196 — all three, as suggested.

## 1. `_currentState` left holding a mid-drag prediction

`_applyDrag` and `_jumpTo` set `_currentState` to `resolveTarget(...)` — a
*prediction* of where the drag would land, not where the sheet is.
`_snapToState` only reconciled it inside `if (state != _settledState)`, so a
drag that ended at the state it started from left the prediction behind: the
sheet rests at `half` while `_currentState` reads `full`.

`evaluateMove` then takes its `currentState == maxState` branch and refuses
every upward drag, so the sheet appears stuck until a downward drag
re-predicts and resyncs it. `onStateChanged` never fires for any of it, so a
host app mirroring sheet state can't detect or work around it.

Reconciling on every snap. Side effects stay gated on `_settledState` as
intended.

## 2. `_onScrollNotification` ignored the notification's axis

It reacted to any `OverscrollNotification` with `overscroll < 0` by claiming
the gesture and re-anchoring `dragStartY` to the finger's current y. Scroll
notifications bubble from every descendant, so a **horizontal** list inside
the sheet emits exactly that shape when pulled past its leading edge.

Two consequences: scrolling sideways dragged the sheet, and the repeated
re-anchoring could hold `dy` under the 10pt threshold indefinitely, leaving
the sheet unable to expand at all.

Now ignores anything whose `metrics.axis` isn't vertical.

## 3. Axis test re-run on every move

`if (dy > threshold && dy > dx)` was evaluated continuously against cumulative
travel from the touch origin, which broke in both directions:

- A sideways swipe grabbed the sheet the instant total vertical travel
  overtook total horizontal.
- After a long sideways scroll, the sheet couldn't be dragged without
  out-travelling that distance vertically — most of the screen on a wide
  swipe.

The `contentDrag` latch compounded it: only ever set from a vertical-looking
move and never re-evaluated, so a horizontal swipe that began with a slight
wobble dragged the sheet for its whole duration.

The axis is now decided **once**, on the first movement past the threshold,
and held until the touch lifts. Below the threshold on both axes the gesture
stays undecided rather than guessing from a few pixels of noise.

## Tests

Three arena cases in `glass_modal_sheet_mechanics_test.dart`:

- a horizontal gesture locks the sheet out for the whole touch
- a long sideways scroll doesn't poison the *next* drag
- sub-threshold movement stays undecided

Plus a widget-level L-gesture in `glass_modal_sheet_state_coverage_test.dart`
— sideways past the leading edge, then upward without lifting. The L-shape is
load-bearing: a pure sideways drag never moved the sheet anyway (the vertical
delta is ~zero), so the finger has to stay down and change direction.

That test moves the sheet **202pt** without these fixes and **3.7pt** with
them, the remainder being `interactionScale` on pointer-down rather than a
drag.

All 148 existing sheet tests pass unchanged.

## Device verification

Built against a physical iPhone (iOS 26.6) with these changes in place, not
just the test harness. The original repro — drag the sheet up short of the
snap threshold, release, then try to expand — no longer sticks, and a
sideways scrub on a horizontal list inside the sheet no longer drags it.

## Note on the full suite

23 tests fail here, all golden/renderer/theme (e.g. `variant: macOS`). They
reproduce on a clean `upstream/main` checkout on this machine, so they're
environmental and unrelated — flagging in case they're green on yours.

## Provenance

Found while debugging a modal sheet over a Mapbox `PlatformView` in an iOS
app. #1 was caught by instrumenting the gesture arena on a physical device —
the state is private, so it took a probe inside the package to see
`currentState/maxState` reading `full/full` while the sheet was rendered at
`half`.

### DIFF
--- a/lib/widgets/overlays/shared/glass_modal_sheet_mechanics.dart
+++ b/lib/widgets/overlays/shared/glass_modal_sheet_mechanics.dart
@@ -533,7 +533,35 @@ class GestureArena {
     final dy = (y - dragStartY).abs();
     final dx = (x - dragStartX).abs();
 
-    if (dy > threshold && dy > dx) {
+    // Decide the axis ONCE per touch, on the first movement past the
+    // threshold, and hold that decision for the rest of the gesture.
+    //
+    // The comparison is still cumulative from the touch origin, but it is now
+    // only ever read at the moment of decision, which fixes two things:
+    //
+    //  * A gesture that started horizontal stays horizontal. Previously the
+    //    test was re-run on every move, so a sideways swipe would grab the
+    //    sheet the instant total vertical travel happened to exceed total
+    //    horizontal — and conversely, after a long sideways scroll the sheet
+    //    could not be dragged at all without out-travelling that distance
+    //    vertically, which on a wide swipe is most of the screen.
+    //  * The reverse latch is gone. `contentDrag` was only ever set from a
+    //    vertical-looking move and then never re-evaluated, so a horizontal
+    //    swipe that began with a slight vertical wobble dragged the sheet for
+    //    its whole duration.
+    //
+    // Below the threshold on both axes the gesture is still undecided; return
+    // false and wait rather than guessing from a few pixels of noise.
+    if (dy <= threshold && dx <= threshold) return false;
+
+    if (dx >= dy) {
+      // Horizontal. An inner scrollable owns this touch; the sheet stays out
+      // of it until the finger lifts and `reset()` clears the phase.
+      phase = GesturePhase.scrolling;
+      return false;
+    }
+
+    if (dy > threshold) {
       // At the topmost detent the sheet can't expand further, so an inner
       // scroll view (if any) owns the gesture. maxState is `full` for a
       // normal sheet but `half` for a half-only sheet (enableFull: false),
--- a/lib/widgets/overlays/shared/glass_modal_sheet_state.dart
+++ b/lib/widgets/overlays/shared/glass_modal_sheet_state.dart
@@ -219,6 +219,15 @@ class _GlassModalSheetState extends State<GlassModalSheet>
     // target mid-gesture. `_settledState` only updates inside this
     // branch, ensuring side effects fire on every consumer-visible
     // state transition.
+    // Reconcile `_currentState` on EVERY snap, not just on consumer-visible
+    // transitions. `_applyDrag` and `_jumpTo` mutate it silently to the
+    // in-flight *predicted* target, so a drag that ends back at the state it
+    // started from leaves that prediction behind: the sheet rests at `half`
+    // while `_currentState` still reads `full`. `evaluateMove` then takes its
+    // `currentState == maxState` branch and refuses every upward drag, so the
+    // sheet appears stuck until a downward drag re-predicts and resyncs it.
+    _currentState = state;
+
     if (state != _settledState) {
       if (state == GlassSheetState.peek || state == GlassSheetState.hidden) {
         HapticFeedback.lightImpact();
@@ -226,7 +235,6 @@ class _GlassModalSheetState extends State<GlassModalSheet>
         HapticFeedback.mediumImpact();
       }
 
-      _currentState = state;
       _settledState = state;
       widget.onStateChanged?.call(state);
 
--- a/lib/widgets/overlays/shared/glass_modal_sheet_state.dart
+++ b/lib/widgets/overlays/shared/glass_modal_sheet_state.dart
@@ -270,6 +270,14 @@ class _GlassModalSheetState extends State<GlassModalSheet>
   // ════════════════════════════════════════════════════════════════════════
 
   bool _onScrollNotification(ScrollNotification notification) {
+    // Vertical only. Scroll notifications bubble from ANY descendant, so a
+    // horizontal list inside the sheet — a date strip, a carousel — emits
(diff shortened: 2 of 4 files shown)


sdegenaar:
Applied manually to main — thanks @jfhair, great fixes as always!!!! Much appreciated.

Cheers

PR #202, opened 2026-08-14
fix(shaders): stop flipping sampling UVs on Impeller GLES (Flutter 3.46+)

### LINKED ISSUE #201: [Bug]: Visual/Rendering regression on Pixel 9 Simulator in Flutter v3.47 (GlassTabbar)
### What's happening?

**Description**:
GlassTabbar fails to render correctly on the Pixel 9 Simulator when using Flutter v3.47.

**To Reproduce**

Run the app on Pixel 9 Simulator using Flutter v3.47.

Navigate to the screen containing GlassTabbar.

Observe the rendering anomaly on the tab bar.

**Expected behavior**
GlassTabbar should render smoothly with expected transparency/blur effects, matching its behavior on iOS Simulators and physical Android devices.

**Environment & Isolation Context**
To help isolate the issue, here is the behavior across different configurations:

🟢 iOS Simulator: Works fine

🟢 Physical Android device: Works fine

🟢 Pixel 9 Simulator (Flutter v3.44): Works fine

🔴 Pixel 9 Simulator (Flutter v3.47): Fails / Rendering bug

This suggests a potential rendering regression introduced between Flutter 3.44 and 3.47 on specific Android Emulator x86_64 / Impeller configurations.

Image

### Steps to Reproduce

Run the app on Pixel 9 Simulator using Flutter v3.47.

Navigate to the screen containing GlassTabbar.

Observe the rendering anomaly on the tab bar.

### Versions

liquid_glass_widgets: ^0.29.5 - Flutter v3.47.

### Platform

Android

### Rendering Backend

None

### Flutter Doctor (optional)

```shell

```

### Anything else?

_No response_

TIANLI0:
Fixes #201.

## What was wrong

Flutter PR [flutter/flutter#186556](https://github.com/flutter/flutter/pull/186556) — *"[Impeller] Retire Y-coord-scale plumbing by flipping GLES at the vertex stage"* — absorbed OpenGL ES's render-to-texture Y-axis difference inside the GLES backend, so render-to-texture content is now stored **top-down on every backend**, matching Metal and Vulkan. It shipped in **Flutter 3.46** and is absent from 3.44/3.45.

Every shader here that samples a Flutter-provided texture still compensates for the *old* bottom-left convention by hand:

```glsl
#ifdef IMPELLER_TARGET_OPENGLES
    screenUV.y = 1.0 - screenUV.y;
#endif
```

On 3.46+ that compensation is no longer a no-op — it actively mirrors every sample about the horizontal centre line. In `liquid_glass_final_render.frag` it hits three places at once: the backdrop UV, the geometry (SDF matte) UV, and the `displacement.y` negation that existed only to re-align refraction with the flipped UV space. So a `GlassTabbar` indicator samples the backdrop from the opposite side of the screen *and* reads its own surface normals upside down.

That explains the otherwise-confusing platform matrix in #201 — physical Android devices run Impeller **Vulkan** and iOS runs **Metal**, neither of which defines `IMPELLER_TARGET_OPENGLES`. Only environments where Impeller falls back to **OpenGL ES** break, which in practice is mostly the Android emulator.

## The fix

New `shaders/gles_compat.glsl` derives the convention once, from `IMPELLER_OPENGLES_UNFLIPPED_DEPRECATED` — the macro `impellerc` defines only on SDKs that removed the flip, added by [flutter/flutter#187316](https://github.com/flutter/flutter/pull/187316) for exactly this migration:

```glsl
#if defined(IMPELLER_TARGET_OPENGLES) && \
    !defined(IMPELLER_OPENGLES_UNFLIPPED_DEPRECATED)
#define LGR_GLES_FLIP_SAMPLE_Y 1
#endif
```

Six sites across four shaders now gate on `LGR_GLES_FLIP_SAMPLE_Y` instead of `IMPELLER_TARGET_OPENGLES`:

| shader | sites |
| --- | --- |
| `liquid_glass_final_render.frag` | backdrop UV, geometry UV, `displacement.y` |
| `interactive_indicator.frag` | refraction offset direction |
| `lightweight_glass.frag` | refraction offset direction |
| `progressive_blur.frag` | backdrop sampling UV |

Because the condition keys off a macro that pre-3.46 toolchains simply don't define, this stays correct on **3.44/3.45 as well as 3.46+** — the `flutter: ">=3.41.0"` constraint in `pubspec.yaml` does not need to move. Keeping the condition in one header also means that when the deprecated macro is eventually removed from Flutter, dropping support for < 3.46 is a one-line change rather than a hunt across five shaders.

I deliberately left `CHANGELOG.md` and the `pubspec.yaml` version untouched — those are yours to decide on release.

## Verification

**Toolchain level.** Probed a shader that `#error`s on the resolved condition against every target `flutter_tools` emits, using the 3.47.0 `impellerc`:

| target stage | `LGR_GLES_FLIP_SAMPLE_Y` |
| --- | --- |
| `--runtime-stage-gles` | not defined ✅ |
| `--runtime-stage-gles3` | not defined ✅ |
| `--runtime-stage-vulkan` | not defined ✅ |
| `--runtime-stage-metal` | not defined ✅ |
| `--sksl` | not defined ✅ |
| GLES with the deprecation macro forced off (simulates ≤ 3.45) | **defined** ✅ |

**Generated-code level.** Compiling before/after with the exact Android target set (`--sksl --runtime-stage-gles --runtime-stage-gles3 --runtime-stage-vulkan`) and normalising SSA temporaries:

- GLES stage: the diff is *only* the removed flips — `_371.y = 1.0 - _371.y`, `_383.y = 1.0 - _383.y`, `_N.y = -_N.y` disappear, nothing else changes.
- SkSL / Metal / Vulkan stages: **identical**. So iOS, physical Android, desktop and web output is provably unaffected by this PR.
- `liquid_glass_geometry_blended.frag`: zero change (it never sampled a Flutter-provided texture).

**On a device.** Android emulator (API 36, x86_64) with the log confirming `Using the Impeller rendering backend (OpenGLES)`, on Flutter 3.47.0. Test page is a `GlassTabBar.bottom` over a vertically asymmetric backdrop — warm red/orange in the top half, cool blue/cyan in the bottom half, with numbered rulers 0–11:

| | what the bar's glass shows |
| --- | --- |
| **before** | warm red/orange and ruler **`0`** — content from the screen *top*, though the bar sits at the bottom over cyan |
| **after** | cyan/teal and ruler **`11`** — the content actually behind it, with refraction and rim lighting intact |

The bar sits at roughly y = 0.95 of screen height; mirrored that is 1 − 0.95 ≈ 0.05, which lands exactly on ruler `0`. Screenshots had to be taken host-side (`adb emu screenrecord screenshot`) — guest `adb shell screencap` returns an all-black frame for Flutter surfaces on this emulator, which is worth knowing if you ever try to reproduce this from a screenshot.

**Suite.** `flutter analyze` clean. `flutter test` shows the same pre-existing failures as an untouched `main` checkout (3 goldens / `liquid_glass_test`), verified by running them on both trees — none related to this change, which is expected given the SkSL output is identical.

## Regression guard

Added `test/shaders/gles_flip_guard_test.dart`. This defect is invisible to the existing suite — it only manifests on the GLES backend, and `flutter test` renders through Skia — so a widget or golden test cannot catch it. Instead it asserts at source level that no shader branches on `IMPELLER_TARGET_OPENGLES` directly, that the compat header keeps both halves of its condition, and that any shader using the guard actually includes the header. Reverting any one of the four shaders makes it fail and print the offending file and line.

## Note on authorship

I used Claude Code to help investigate and draft this fix — locating the upstream Flutter change that caused the regression, and building the compile-time and on-device verification described above. I have reviewed the diff, and the reproduction and before/after behaviour were checked by hand on a real emulator. Flagging it in the interest of the repo's AI contribution guidelines; happy to rework anything that doesn't match how you'd like this handled.

### DIFF
--- a/docs/PROGRESSIVE_BLUR.md
+++ b/docs/PROGRESSIVE_BLUR.md
@@ -63,8 +63,11 @@ The bound texture is the **whole backdrop** (the screen), not the bar's own
 rectangle, so `uSize` is the screen size. To make the gradient run across the
 *bar* rather than the whole screen, the widget passes its own device-pixel
 rectangle (`uRegionOriginPx` / `uRegionSizePx`) into the shader and normalises
-the gradient over that. The GLES backend captures the backdrop y-flipped, which
-the shader corrects under `IMPELLER_TARGET_OPENGLES`.
+the gradient over that. On Flutter 3.44/3.45 the GLES backend captured the
+backdrop y-flipped, which the shader corrects under `LGR_GLES_FLIP_SAMPLE_Y`
+(see `shaders/gles_compat.glsl`). Flutter 3.46+ stores render-to-texture content
+top-down on every backend, so the correction is compiled out there — applying it
+anyway is what caused #201.
 
 ## Usage
 
--- a/shaders/gles_compat.glsl
+++ b/shaders/gles_compat.glsl
@@ -1,0 +1,60 @@
+// Copyright 2026, Sebastian Degenaar for pixel-innovations.com (liquid_glass_widgets)
+//
+// SPDX-License-Identifier: MIT
+//
+// OpenGL ES texture-origin compatibility.
+//
+// ## Background
+//
+// Impeller's GLES backend used to store render-to-texture content with GL's
+// native bottom-left origin, while Metal and Vulkan stored it top-down. Every
+// fragment shader that sampled a Flutter-provided texture (a BackdropFilter
+// snapshot, a `toImageSync` capture, a sampler bound via `setImageSampler`) had
+// to compensate by hand:
+//
+//     #ifdef IMPELLER_TARGET_OPENGLES
+//         uv.y = 1.0 - uv.y;
+//     #endif
+//
+// Flutter PR #186556 ("[Impeller] Retire Y-coord-scale plumbing by flipping
+// GLES at the vertex stage") absorbed that difference inside the GLES backend,
+// so render-to-texture content is now stored top-down on *every* backend. The
+// manual flip above became not merely unnecessary but actively wrong: it
+// mirrors every sample vertically.
+//
+// That change shipped in **Flutter 3.46**. It is absent from 3.44 and 3.45.
+//
+// ## How to detect which convention the toolchain uses
+//
+// `impellerc` defines `IMPELLER_OPENGLES_UNFLIPPED_DEPRECATED` on the GLES
+// runtime stages — and *only* on engines that have removed the flip (Flutter
+// PR #187316 added it for exactly this purpose). Its presence therefore means
+// "textures are already top-down; do not flip". Its absence on a GLES stage
+// means we are being compiled by Flutter 3.44/3.45, which still needs the flip.
+//
+// [LGR_GLES_FLIP_SAMPLE_Y] encodes that decision once. Shaders sampling a
+// Flutter-provided texture must gate their Y compensation on it rather than on
+// `IMPELLER_TARGET_OPENGLES` directly:
+//
+//     #ifdef LGR_GLES_FLIP_SAMPLE_Y
+//         uv.y = 1.0 - uv.y;
+//     #endif
+//
+// Keeping the condition in one header means the eventual removal of the
+// deprecated macro from Flutter (at which point the package can drop support
+// for < 3.46) is a one-line change here, not a hunt across five shaders.
+//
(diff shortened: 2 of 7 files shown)


sdegenaar:
Hey @TIANLI0, thank-you very much for the PR. Had started looking at the this, but your solution was better. Nice work!

Cheers

PR #211, opened 2026-08-19
fix(ProgressiveBlur): resolve the shader region at paint time

### LINKED ISSUE #210: ProgressiveBlur: shader region origin is hard-coded to (0,0), so it only works at the top-left
`ProgressiveBlur` renders correctly only when it sits at the top-left of its backdrop layer. Anywhere else, the gradient is normalised over the wrong rectangle.

The shader already handles this — `progressive_blur.frag` declares `uniform vec2 uRegionOriginPx` (uniforms 6/7) and computes:

```glsl
(FlutterFragCoord().xy - uRegionOriginPx) / uRegionSizePx
```

It's `progressive_blur.dart` that discards it:

```dart
..setFloat(6, 0) // region origin x (device px)
..setFloat(7, 0) // region origin y (device px)
```

with the note *"The bar is anchored at the top-left of the backdrop layer (true for top app bars)"*. That holds for the intended use, but not for a blur band inside a modal sheet, a bottom bar, or any inset container.

### What we saw

iOS 26.6, iPhone 16 Pro Max, Impeller, `ImageFilter.isShaderFilterSupported == true` throughout — so not a backend fallback:

- Used as-is inside a sheet's top edge → the sheet renders **black**.
- Wrapped in a `RepaintBoundary` to try to make `(0,0)` true → the filter samples an empty layer, so **no blur at all**, silently.

Both still reproduce on 0.29.7.

### Suggested fix

Pass the widget's own offset in the backdrop's coordinate space instead of `(0,0)`:

```dart
final origin = (context.findRenderObject() as RenderBox).localToGlobal(Offset.zero);
..setFloat(6, origin.dx * dpr)
..setFloat(7, origin.dy * dpr)
```

The shader needs no change. Happy to open this as a PR if you'd like it.

### Why it matters beyond our case

`GlassScrollEdgeEffect` fades but doesn't blur, so `ProgressiveBlur` is the package's only graduated-blur primitive — and today it can only be used at the top-left.

Ours is a scroll edge inside a sheet over a Mapbox `PlatformView`, where the single-pass shader is worth a great deal. Replacing six stacked `BackdropFilter`s with one pass took that screen from **20.8 ms sustained raster / 30 fps / 86% of frames late** to **2.4 ms / 121 fps / zero late frames out of 596**. Every backdrop overlapping a platform view costs an extra overlay composite and they compound — the same widget runs twelve slices for 1.8 ms on a screen with no platform view under it.

Thanks for the 0.29.7 progressive-blur work, by the way — the halved texture reads are exactly the direction that made this worth building on.

jfhair:
Fixes #210.

`ProgressiveBlur` hard-coded its shader region origin to `(0, 0)`:

```dart
..setFloat(6, 0) // region origin x (device px)
..setFloat(7, 0) // region origin y (device px)
```

The shader normalises its gradient over that rectangle, so this is correct only
where the note said it was — a bar anchored at the top-left of the backdrop
layer. Anywhere else — a blur band inside a modal sheet, or any inset container
— the gradient is computed over the wrong rect: in practice the surface renders
black, and wrapping it in a `RepaintBoundary` to try to make `(0, 0)` true
instead samples an empty layer, so the blur silently disappears.

### The fix

Resolve the region in a `RenderObject` at paint time rather than in `build`.

Paint is the only place the answer is both knowable and current. The widget's
offset within the backdrop layer changes whenever an ancestor **moves** it — a
sheet being dragged, a scroll, an animated inset — and those move it by
repainting, not by rebuilding. Reading the offset during `build`, or in a
post-frame callback, leaves the gradient normalised over a stale rectangle for
as long as the motion lasts. That was worth getting right for us specifically:
ours is a scroll edge inside a draggable sheet.

Dropping the `LayoutBuilder` falls out of the same change, and has a small bonus
— a `ProgressiveBlur` can now sit under a parent that asks for intrinsic
dimensions, which `LayoutBuilder` refuses to answer.

No shader change; `progressive_blur.frag` already declared `uRegionOriginPx` and
used it.

### Tests

The uniform arithmetic moves into `progressiveBlurUniforms`, a pure function,
and is now covered — including that a non-zero origin survives, which is the
regression itself.

This seam is deliberate. The defect was a literal `0` in the middle of a widget
build, and no test could have caught it: the shader path needs a compiled
`FragmentProgram`, which a headless VM never provides, so those lines are
`coverage:ignore` and the existing tests necessarily exercise the uniform-blur
fallback. Pulling the region maths out gives that part of the path something a
test can hold onto.

`flutter test` is green apart from five golden tests that also fail unchanged on
`origin/main` here (`container_widgets`, `overlay_widgets`,
`surfaces/glass_toolbar`) — machine-dependent goldens, untouched by this change.

### Verification

On device — iOS 26.6, iPhone 16 Pro Max, Impeller, `isShaderFilterSupported ==
true` — in the exact configuration from the report: a scroll-edge blur inside a
sheet over a Mapbox `PlatformView`. It rendered black before; it now blurs
correctly and the gradient tracks the sheet through a drag between detents.

### DIFF
--- a/lib/widgets/effects/progressive_blur.dart
+++ b/lib/widgets/effects/progressive_blur.dart
@@ -1,5 +1,6 @@
 import 'dart:ui' as ui;
 
+import 'package:flutter/rendering.dart';
 import 'package:flutter/widgets.dart';
 
 /// The edge a [ProgressiveBlur] is *strongest* at; it eases to perfectly sharp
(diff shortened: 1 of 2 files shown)


jfhair:
Codecov's flag was fair, and half of it was mine.

The shader branch in `build()` was already `coverage:ignore`-marked before this
change, for the reason documented there: it needs a compiled `FragmentProgram`,
which a headless VM never provides. My rewrite moved that code into new classes
and dropped the markers on the way. Restored in `ef9c5a7`.

The remainder was the new render object, unreachable for the same reason. Rather
than just marking it, I shrank what needs marking — the six per-field setters
collapse into a single `update()`, since every field arrives from the same widget
on the same rebuild, so one comparison is all the change detection this needs.
That removed about 40 lines of unreachable ceremony at roughly break-even on
total size.

Uncovered lines in the file: **51 → 2**. Both survivors are the pre-existing
`debugPrint` on the shader-load failure path, unchanged from `main`. The patch
now adds none.

On the broader point, since a rendering fix landing as mostly ignore-marked code
is a reasonable thing to be uneasy about: that's precisely why the region
arithmetic came out into `progressiveBlurUniforms` as a pure function. The
original defect was a literal `0` sitting in the middle of a widget build, in a
region no test could reach — so the fix that matters most for regressions is that
the maths now lives somewhere a test can hold onto, including a case asserting a
non-zero origin survives. The part that remains uncovered is layer plumbing that
only a GPU can execute.

sdegenaar:
Thanks again @jfhair. Really appreciate your contributions! Will be in v0.30.1 released shortly.

Cheers

Issue #214, opened 2026-08-21
[Bug]: GlassPullDownButton crashes/fails when quality drops to minimal (InheritedGeometryRenderLink missing)


minhtritc97:
### What's happening?

When using GlassPullDownButton, GlassMenu always wraps its content in a LiquidGlassBlendGroup. However, when the performance/rendering quality falls back to minimal, AdaptiveLiquidGlassLayer skips creating a LiquidGlassLayer.

As a result, there is no InheritedGeometryRenderLink available in the context for the LiquidGlassBlendGroup to attach to, causing an exception/rendering failure.

### Expected Behavior
GlassPullDownButton (or GlassMenu) should handle low/minimal quality gracefully when LiquidGlassLayer (and its InheritedGeometryRenderLink) is omitted, either by disabling the blend group requirement or providing a fallback.

### Actual Behavior:
The widget fails because LiquidGlassBlendGroup expects an active InheritedGeometryRenderLink from the parent layer, which is bypassed in minimal quality mode.

### Steps to Reproduce

1. Use GlassPullDownButton in a view with LiquidGlassBlendGroup.

2. Trigger or set the device/layer quality mode to minimal (where AdaptiveLiquidGlassLayer bypasses the LiquidGlassLayer).

3. Open or render the GlassPullDownButton menu.

### Versions

liquid_glass_widgets: 0.30.1 Flutter: 3.47.0

### Platform

Android

### Rendering Backend

None

### Flutter Doctor (optional)

```shell

```

sdegenaar:
Hey @minhtritc97, thanks for reporting this and the analysis work. Fixed in v0.30.2.

Let me know how you go.

Cheers

minhtritc97:
> Hey [@minhtritc97](https://github.com/minhtritc97), thanks for reporting this and the analysis work. Fixed in v0.30.2.
> 
> Let me know how you go.
> 
> Cheers

The bug is resolved, thanks for your quick response

PR #217, opened 2026-08-25
feat(GlassModalSheet): morph presentation from a trigger button

### LINKED ISSUE #216: [Feature]: Morph GlassModalSheet from a trigger button
### What kind of feature?

Animation or interaction improvement

iOS 26's native Liquid Glass supports morphing a glass button into a presented surface (glassEffectID / GlassEffectContainer - the trigger empties, a glass droplet detaches, inflates while travelling, and lands as the sheet; dismissal reverses back into the button).

GlassMenu already implements exactly this pattern via the Liquid Morph Engine, and `LIQUID_MORPH_ENGINE.md` notes the engine is decoupled so future widgets can consume it. GlassModalSheet, however, still presents with a SlideTransition from the screen edge (the fillTransition width-morph is the closest it gets).

Let GlassModalSheet (or showGlassModalSheet) present via the morph engine from a caller-supplied trigger — e.g.:

```dart
GlassModalSheet.show(
  context,
  morphFrom: triggerKey,          // GlobalKey or Rect of the trigger
  morphSpeed: MorphSpeed.normal,  // existing engine vocabulary
  ...
);
```

with the sheet's detents/drag-dismiss unchanged once settled, and drag-dismiss below the last detent handing back to the engine's close morph.

A single-item GlassMenu with menuWidth/menuHeight shaped like a sheet gets close, but the height estimation assumes 44px for non-GlassMenuItem children, the selection gestures compete with inner scrollables, and there's no edge docking or drag-to-dismiss - so a first-class sheet consumer of the engine would be the real solution.

### Anything else?

https://github.com/user-attachments/assets/77d743f9-bc65-480e-bdc8-b206619bbc95

JakeThomson:
Fixes #216 - the second consumer of the Liquid Morph Engine, which
`LIQUID_MORPH_ENGINE.md` anticipated when it noted the engine is decoupled so future widgets can use it.

## What changed

`GlassModalSheet.show()` takes a trigger and presents by morphing out of it
instead of sliding up from the edge: the trigger empties, a glass droplet
detaches and inflates along the J-curve, and lands as the sheet. Dismissing
pours it back.

https://github.com/user-attachments/assets/dfeb2f7c-104e-41a0-81b2-2a8c90978ffb

```dart
GlassMorphTrigger(
  builder: (context, anchor) => GlassButton(
    onTap: () => GlassModalSheet.show(
      context: context,
      morphFrom: anchor,
      builder: (context) => const MySheetBody(),
    ),
    child: const Icon(CupertinoIcons.add),
  ),
)
```

| parameter | type | effect |
|---|---|---|
| `morphFrom` | `GlassMorphAnchor?` | full morph — the trigger empties and the teardrop neck stretches |
| `morphFromRect` | `Rect?` | escape hatch for a trigger that can't be wrapped; blooms from a point |
| `morphSpeed` | `MorphSpeed` | reuses the engine's existing vocabulary, default `normal` |

Everything after the morph is untouched — detents, drag, snap, the controller.
Only the presentation changes.

## Why the trigger needs a wrapper rather than a GlobalKey

My first cut took a `GlobalKey`, which is enough to *locate* the trigger and aim
the droplet. It looked wrong on device, and the reason is structural: nothing in
Flutter lets one widget hide another it doesn't own, so the trigger kept painting
underneath while a glass droplet inflated on top of it. That reads as a
duplicated button, not a morph — worst in `standard`, where the rims are crisper.

`GlassMenu` doesn't hit this because it owns its trigger and can set it to
`Opacity(0)` for the duration. `GlassMorphTrigger` makes that same arrangement
available to a trigger the sheet doesn't own: it owns the key *and* the opacity,
and `GlassMorphAnchor` is the channel back from the presented route.

`morphFromRect` stays for triggers that genuinely can't be wrapped. Since the
trigger keeps painting there, the anchor blob is suppressed and the droplet
blooms from the rect's centre — no second button-sized shape at any point, at
the cost of the teardrop neck, which needs two blobs to stretch between. Same
trade-off as `GlassMenu.morphFromZero`.

## Why the droplet isn't the sheet itself

The teardrop neck is drawn by the SDF metaball shader, which only bridges shapes
inside one glass layer's blend group — and the real sheet owns its own
`AdaptiveGlass` layer, so it structurally cannot merge with the trigger ghost.
The morph therefore renders its own two-blob droplet and hands off to the sheet
at the settled instant, when both are motionless on the identical frame.

The sheet stays mounted underneath for the whole morph and is *revealed*, never
inserted — it keeps its slot in the same `Stack` throughout, because remounting a
glass widget mid-animation re-seeds its layers and springs and shows as a glitch
frame. That was worth being careful about; an earlier version returned the sheet
bare once settled, which changed its depth and tore the subtree down at exactly
the moment the handoff is meant to be invisible.

Aiming is not a parallel calculation: `SheetMorphGeometry.restingRect` derives
the destination from `SheetGeometry.positionForState`, the same call the sheet's
own `_calculateMetrics` uses, so the droplet and the sheet cannot drift apart.
It also mirrors the resting surface per detent, so a glass `medium` hands off to
glass and an opaque `large` hands off to colour.

## Why the route duration is sized the way it is

The route reverse is the window the closing morph plays in, and I got this wrong
twice before measuring it. For `MorphSpeed.normal` the droplet is caught at
**304 ms** but the spring doesn't fully settle until **696 ms**.

Sizing the route to the full settle keeps the modal barrier up for that whole
time, and the barrier swallows every touch — so the trigger sits there visibly
back but dead for ~400 ms. `GlassMenu` drops its own barrier 30 % into the close
precisely so that doesn't happen, and you can grab its trigger mid-animation.

So the route is sized to the droplet's journey and no further (380 ms at
`normal`), and `GlassMorphTrigger` continues the spring on its own ticker from
the value and velocity it's handed at the catch. The bounce finishes after the
route is gone, and the button is tappable throughout — grabbing it cancels the
bounce and reopens.

One deviation worth naming: that tail always runs the native-parity spring rather
than the per-speed one, because `GlassMorphController`'s spring mapping is
private. Over a bounce this small it's sub-pixel, and it saves the trigger from
having to know the speed — but say the word and I'll expose the mapping instead.

## Back-compat

Purely additive. Leave `morphFrom` and `morphFromRect` null and the existing
`SlideTransition` presentation is unchanged — same duration, same curve, same
route. No removals; existing call sites compile untouched.

The morph is skipped wholesale wherever the metaball blend is unavailable, and
the sheet presents with the ordinary slide instead: Skia/web
(`ImageFilter.isShaderFilterSupported == false`), `GlassQuality.minimal`, and
`platformViewBackdrop` — the same cases that skip `LiquidGlassBlendGroup` per
#214. A morph without the neck is two glass shapes and no bridge, which reads
worse than the slide it would replace, so degrading seemed wrong.

Reduce Motion is honoured by the engine's instant spring. The presenter opens
from `didChangeDependencies` rather than `initState` for that reason — the flag
has to reach the controller before the spring starts, or the first presentation
of every session animates at full length with Reduce Motion on.

## Tests & coverage

Effective project coverage **91.99%** (gate is 90%). Patch coverage **99.7%**
(305/306 added executable lines).

59 new tests: the pure geometry (resting frames per detent, blob interpolation,
radius, fill opacity), `GlassMorphTrigger` (empties and restores, runs its bounce
and returns to exactly the resting position, a fresh open cancels an in-flight
bounce), the presenter (handoff latch, the sheet element surviving the handoff,
reduce motion, the closing gate), the route wiring, and every fallback.

**One added line is uncovered** — `SheetMorphGeometry._()`, the private
constructor of a pure utility class, which is never invoked. `LiquidMorphPhysics`
has exactly the same uncovered line for the same reason.

## Verification

On the iOS 27 simulator (iPhone 17, Impeller, `isShaderFilterSupported == true`)
via `example/integration_test/morph_demo_test.dart`, which drives the real demo
and asserts the morph route is taken rather than the fallback — through the open,
the landing, and the barrier-tap close.

Not tested on a physical device, on Android, or on desktop. The Skia fallback is
covered by widget tests rather than a browser run.

`flutter analyze --fatal-warnings` clean; `dart format` clean.

## Two things to say no to if you'd rather

- **`example/integration_test/morph_demo_test.dart`** adds an `integration_test`
  dev-dependency to the example, and needs a platform folder generated to run,
  which the repo doesn't ship. It exists because a headless run reports no
  Impeller, so `show()` always takes the slide fallback there and the morph path
  is otherwise unverifiable. Happy to drop it.
- **`GlassModalSheet.debugMorphSupportsBlending`** is a `@visibleForTesting`
  override standing in for that same probe, so the route wiring is still tested
  headlessly. `platformViewBackdrop` and `minimal` continue to disable the morph
  under it, so those fallbacks stay honest. Also happy to drop it, at the cost of
  that coverage.

## Notes

- `CHANGELOG.md` is under `# Unreleased` — move it under whatever version you're
  cutting next.
- Docs updated: README feature bullet, a consumers table and a section on what
  differs when the destination is a route in `LIQUID_MORPH_ENGINE.md`, and the
  three new parameters plus a "Morphing from a Trigger" section in
  `GLASS_MODAL_SHEETS_GUIDE.md`.
- `SheetMorphGeometry` and `GlassSheetMorphPresenter` are public within the
  library but deliberately not exported from `liquid_glass_widgets.dart`, the way
  `SheetGeometry` and `GestureArena` already are — reachable from the tests that
  import the library file directly, without widening the package's public
  surface. Only `GlassMorphTrigger` and `GlassMorphAnchor` are exported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+# Unreleased
+
+## Features
+
+- **`GlassModalSheet` morphs from a trigger button (#216):** `show()` gains `morphFrom` — the `GlassMorphAnchor` from a new `GlassMorphTrigger` wrapper — presenting the sheet with the iOS 26 liquid morph instead of the slide-up: the trigger empties, a glass droplet detaches and inflates as it travels, and lands as the sheet. Makes the sheet the second consumer of the Liquid Morph Engine after `GlassMenu`. `morphSpeed` tunes the spring.
+- **`GlassMorphTrigger` / `GlassMorphAnchor` added:** the wrapper owns the trigger's key *and* its opacity, so the trigger can empty itself for the morph and take the closing bounce on its own ticker. A `GlobalKey` cannot hide a widget it doesn't own, and a trigger still painting under the droplet reads as a duplicated button. `morphFromRect` remains for triggers that can't be wrapped, blooming from a point.
+- **Slide presentation untouched:** leave `morphFrom` / `morphFromRect` null and nothing changes. The morph falls back to the slide wherever the metaball blend is unavailable — Skia/web, `GlassQuality.minimal`, `platformViewBackdrop` — and Reduce Motion resolves it through the engine's instant spring.
+
+---
+
 # 0.30.2
 
 ## Bug Fixes
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Both parameters are optional — omit them and the library uses sensible default
 ## Features
 
 - **Comprehensive glass widget library** — containers, interactive controls, inputs, feedback, overlays, and navigation surfaces (see [Widget Categories](#widget-categories))
-- **Liquid Morph Engine** — a standalone physics system powering iOS 26-style liquid morphing. `GlassMenu` is the first consumer; future widgets will use the same engine for consistent liquid transitions. See [`docs/LIQUID_MORPH_ENGINE.md`](docs/LIQUID_MORPH_ENGINE.md)
+- **Liquid Morph Engine** — a standalone physics system powering iOS 26-style liquid morphing. `GlassMenu` morphs out of its trigger button, and `GlassModalSheet.show(morphFrom:)` presents a modal sheet out of a `GlassMorphTrigger` the same way — the trigger empties, a glass droplet inflates as it travels, and dismissing pours it back. See [`docs/LIQUID_MORPH_ENGINE.md`](docs/LIQUID_MORPH_ENGINE.md)
 - **Real frosted glass** — native two-pass Gaussian blur + shader refraction on Impeller; lightweight shader on Skia/Web
 - **Just works everywhere** — iOS, Android, macOS, Web, Windows, Linux; rendering path chosen automatically
 - **Adaptive quality** *(experimental)* — `GlassAdaptiveScope` benchmarks the device at startup and adjusts quality in real time: `minimal` on slow hardware, `standard` on mid-range, `premium` on fast devices. Degrades on thermal throttle, recovers when cool
--- a/docs/GLASS_MODAL_SHEETS_GUIDE.md
+++ b/docs/GLASS_MODAL_SHEETS_GUIDE.md
@@ -20,6 +20,9 @@ Settings for the primary content and the sheet's basic behavior.
 | `isDismissible` | `bool` | `true` | (In `show`) Controls `barrierDismissible`. If `true`, the `Scaffold` background triggers `snapToState(SheetState.hidden)`. |
 | `useRootNavigator` | `bool` | `false` | (In `show`) If true, pushes the modal onto the top-level `Navigator`, bypassing local navigation stacks. |
 | `barrierLabel` | `String` | `'Dismiss'` | Accessibility label for the modal barrier. |
+| `morphFrom` | `GlassMorphAnchor?` | `null` | (In `show`) Presents by morphing out of a `GlassMorphTrigger` instead of sliding up. The anchor both locates the trigger and empties it for the duration. |
+| `morphFromRect` | `Rect?` | `null` | (In `show`) For triggers that can't be wrapped — an explicit global rect. The trigger stays painted, so the anchor blob is suppressed and the droplet blooms from the rect's centre. Mutually exclusive with `morphFrom`. |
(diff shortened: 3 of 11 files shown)


sdegenaar:
Hey @JakeThomson, really appreciate the two PR's. Quality work and hitting the gaps. Noticed another 2 PR's today so excited to get into them. Again, really appreciate it and the great contributions you are making.

Cheers

JakeThomson:
No worries @sdegenaar! Thanks for all your hard work on putting this package together, Flutter has been lacking a decent implementation of liquid glass for too long now :)

Issue #198, opened 2026-08-11
[Bug]: title no center


Vincen-dev:
### What's happening?

<img width="636" height="71" alt="Image" src="https://github.com/user-attachments/assets/31792063-5192-4eff-8c09-5be1dfc23f8f" />

### Steps to Reproduce

add backbutton  title no center

### Versions

0.29.2 + 3.44.3

### Platform

iOS

### Rendering Backend

None

### Flutter Doctor (optional)

```shell

```

sdegenaar:
Hi @Vincen-dev,
Thanks for the report! Could you share a small code snippet showing how you've set up your navigation bar and back button? That'll help us reproduce the issue quickly.

Cheers

Vincen-dev:
class EvoAppBar extends GlassAppBar {
  const EvoAppBar({
    super.key,
    super.title,
    super.leading = const _EvoAppBarBackButton(),
    super.actions,
    super.centerTitle = true,
    super.backgroundColor = const Color(0x00000000),
    super.toolbarHeight = 44.0,
    super.padding = EvoInsets.pageHorizontal,
    super.buttonSettings,
    super.largeTitleController,
    super.bottom,
  });

  factory EvoAppBar.withBackAction({Key? key, required VoidCallback onBack}) =>
      EvoAppBar(
        key: key,
        leading: _EvoAppBarBackButton(onTap: onBack),
      );
}

final class _EvoAppBarBackButton extends StatelessWidget {
  const _EvoAppBarBackButton({this.onTap});

  final VoidCallback? onTap;

  @override
  Widget build(BuildContext context) {
    final colors = EvoThemeResolver.colorSchemeOf(context);
    return GlassButton(
      label: CupertinoLocalizations.of(context).backButtonLabel,
      width: 36,
      height: 36,
      iconSize: 20,
      iconColor: colors.labelPrimary,
      stretch: 0.15,
      icon: const Icon(CupertinoIcons.back),
      onTap: onTap ?? () => unawaited(Navigator.maybePop(context)),
    );
  }
}

@sdegenaar just add leading

sdegenaar:
Hey @Vincen-dev, thanks for the sample code. Fixed in v0.29.5 

Let me know how you go.

Cheers

sdegenaar:
Closing this as stale — feel free to reopen if this is still an issue in the latest version

Cheers

PR #221, opened 2026-08-25
feat(GlassNavigationShell): pin navigation-bar glass chrome across route transitions

### LINKED ISSUE #220: [Feature]: GlassNavigationTransition - pinned nav-bar chrome across route transitions
### What kind of feature?

New widget

On iOS 26, navigation bar items belong to the navigation *stack*, not to any one screen. During a push/pop (including the interactive back-swipe), the page content and title slide with the transition, but the glass back button and the trailing actions capsule stay pinned in place — and the capsule morphs in place into the next screen's actions (icons cross-fade, the capsule resizes, matched items hold their position). You can see this in any iOS 26 app; the GitHub app is a good reference.

`liquid_glass_widgets` can't reproduce this today: `GlassScaffold` renders `GlassAppBar` inside each route's own subtree, so the whole bar — buttons included — slides with the page. This is already on the roadmap as `GlassNavigationTransition` (docs/ROADMAP.md, "coordinated glass morphing during CupertinoPageRoute push/pop transitions"), I am not sure if this solution covers everything originally intended by that item.

| Native | Current |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/779c1465-3145-440e-a410-ee25b2bfb555" /> | <video src="https://github.com/user-attachments/assets/03063f28-c60d-488f-aff5-79cb83089023" />  | 

Apple documents the native behaviour on [`UIBarButtonItem.identifier`](https://developer.apple.com/documentation/uikit/uibarbuttonitem/identifier): *"When the set of bar button items in a navigation bar or toolbar changes (for example, when pushing or popping view controllers), UIKit automatically animates the transition between the different sets of items. By default, UIKit uses heuristics based on item position and content to determine which items should be matched."*

A navigator-level host plus a data-driven item API on `GlassAppBar`, mirroring `UIBarButtonItem`.

**One-line install**, wrapping whatever Navigator the app builds:

```dart
CupertinoApp(
  builder: (context, child) => GlassNavigationShell(child: child!),
  home: const HomeScreen(),
);
```

**Screens declare items as data:**

```dart
GlassAppBar(
  title: const Text('Repository'),
  pinnedActions: [
    GlassBarItem.icon(icon: const Icon(CupertinoIcons.add), id: 'add', onTap: _add),
    GlassBarItem.icon(icon: const Icon(CupertinoIcons.ellipsis), onTap: _more),
  ],
)
```

Key semantics, all mirroring native:

- **Back button is automatic** — shown iff the route can pop (`ModalRoute.impliesAppBarDismissal`), never on a root; default action `Navigator.maybePop()`, with an `onBack` override for router-specific pops (`context.pop()`).
- **`id` mirrors `UIBarButtonItem.identifier`** — items sharing an id across routes are treated as the same item and stay put while everything around them morphs; unidentified items match positionally from the trailing edge (UIKit's documented default heuristic).
- **`GlassBarItem.custom(child: ...)` is the `customView:` analogue** — any widget, measured at intrinsic width during layout so the capsule sizes itself around it. Deliberately *no* API to offset/reposition clusters, because iOS 26 has none either.
- **Router-agnostic** — the shell only reads `ModalRoute` animations (no NavigatorObserver, no navigation interception), so it works identically with the imperative Navigator and any Pages-API router (go_router, auto_route, beamer), installed via `CupertinoApp.router(builder: ...)`.
- **Scrubbable** — everything is a pure function of the top route's animation value, so the interactive back-swipe scrubs the morph proportionally and cancelled swipes rebound for free.
- **Graceful fallback** — with no shell present (or `GlassQuality.minimal`), the same `pinnedActions` data renders in-route as an automatic back `GlassButton` + `GlassButtonGroup.icons` capsule, so screens work either way.
- **Respects the package's glass rules** — glass opacity is never animated (only geometry animates; only item contents cross-fade), and the pinned chrome lives in its own isolated layer per the post-0.14.0 backdrop-isolation rules.

- **Counter-translating the in-route bars** (applying the inverse of the route's slide so bars appear pinned): dead on arrival — it requires two overlapping, partially-faded glass pills mid-transition, and partially fading a glass surface pops its backdrop pass (the rule documented in the modal-sheet morph code).
- **Morph hand-off** (GlassMorphTrigger-style hide/stand-in/hand-back): viable but adds a pixel-perfect swap-instant problem the persistent host simply doesn't have, especially at back-swipe start.
- The navigator-level host turned out to be both the simplest and the most faithful option — it's literally how UIKit structures it.

### Anything else?

I see this is on the roadmap so I have had a crack at the implementation of all of the above on my fork, verified on the iOS simulator with Impeller: pinned glass correctly refracts the sliding page every frame, identifier matching works, custom widgets measure correctly, and the go_router demo behaves identically to the imperative-Navigator demo. I'll create a draft PR to support the discussion: https://github.com/sdegenaar/liquid_glass_widgets/pull/221

JakeThomson:
Fixes #220 — the roadmap's `GlassNavigationTransition`, which ROADMAP.md sketches as
"coordinated glass morphing during CupertinoPageRoute push/pop transitions".

## What changed

| Native | Before | After |
|--------|--------|--------|
| <video src="https://github.com/user-attachments/assets/779c1465-3145-440e-a410-ee25b2bfb555" /> | <video src="https://github.com/user-attachments/assets/03063f28-c60d-488f-aff5-79cb83089023" />  | <video src="https://github.com/user-attachments/assets/aaf7e706-9256-4d57-91bc-e44877586ec4" /> | 

On iOS 26 the bar belongs to the navigation stack, not the screen: during a push/pop the
page and title slide while the back button and actions capsule stay pinned, and the capsule
morphs in place into the next screen's items. Apple states it on
[`UIBarButtonItem.identifier`](https://developer.apple.com/documentation/uikit/uibarbuttonitem/identifier):
*"When the set of bar button items in a navigation bar or toolbar changes (for example, when
pushing or popping view controllers), UIKit automatically animates the transition between the
different sets of items."* This PR is that model for `GlassScaffold` screens.

One line installs it, and screens opt in with a constructor — the
`GlassButtonGroup` / `GlassButtonGroup.icons` widgets-vs-data pattern:

```dart
CupertinoApp(
  builder: (context, child) => GlassNavigationShell(child: child!),
  home: const HomeScreen(),
);

GlassAppBar.pinned(
  title: const Text('Repository'),
  actions: [
    GlassBarItem.icon(icon: const Icon(CupertinoIcons.add), id: 'add', onTap: _add),
    GlassBarItem.menu(
      icon: const Icon(CupertinoIcons.ellipsis),
      menuItems: [GlassMenuItem(title: 'Share', onTap: _share)],
    ),
  ],
)

// The common back-button-only screen collapses to:
GlassAppBar.pinned(title: Text('Surfaces'))
```

| surface | role |
|---|---|
| `GlassNavigationShell` | hosts the pinned chrome above the `Navigator`; installed once via the app `builder` (plain or `.router`) |
| `GlassAppBar.pinned(title:, actions:, backButton:, onBack:)` | the participating mode — actions as data, automatic back button off-root defaulting to `Navigator.maybePop()`. The plain `GlassAppBar` keeps the widget `leading`/`actions` API and never participates; the constructor choice is the mode, so the two APIs cannot be mixed on one bar |
| `GlassBarItem.icon / .menu / .custom / .spacer` | `UIBarButtonItem` analogues — data, not widgets; `.menu` is `UIBarButtonItem.menu`, `.custom` is `customView:`, `.spacer` is `fixedSpace` (parsed, rendering is a follow-up) |
| `id` | `UIBarButtonItem.identifier` — same id across routes ⇒ same item, it holds position while the rest morphs; otherwise UIKit's trailing-edge positional heuristic |

The title stays in-route and slides with the page — that matches native, where large titles
scroll with content and only the item clusters pin.

The example app adopts it across the board: the showcase installs the shell, every category
page pins its back button (`GlassAppBar.pinned()`), the nav-patterns sub-demos declare their
actions as data, and a new **Nested Navigation** pattern walks a three-level drill-down with
an identifier-matched item holding position at each push. Net example diff is strongly
negative — the data API replaces a lot of repeated hand-built back-button code.

Two patterns are the ones to look at: **Nested Navigation**, and **Bar Item Types**, which
puts all three item kinds in one capsule — a custom unread badge, a plain icon and a
pull-down — then pushes to a second screen where the badge (same `id` on both) holds
position while the capsule widens around it, 3 → 128, and the icon leaves. Nothing about
that width is hardcoded; the capsule measures the badge.

## Where this is heading

**`.pinned` is a transition vehicle, not the destination.** On iOS 26 the pinned behaviour
*is* the navigation bar — it isn't a mode you opt into, it's what a navigation bar does. So
the intended trajectory, now recorded in `ROADMAP.md` and the docs guide, is:

1. **This release** — land the data API additively. Nothing shipped changes: the plain
   `GlassAppBar` constructor is untouched and never participates in pinning.
2. **Next** — bring `GlassBarItem` to parity with the widget API: menu items (`GlassMenu`
   pill triggers), spacers / multi-capsule grouping, text and prominent styles. Until that
   parity exists, forcing anyone onto the data API would take away things they can express
   today, which is why this lands opt-in rather than as the default.
3. **At a major of your choosing** — make pinning the default `GlassAppBar`, demoting the
   widget-based `leading`/`actions` constructor to a legacy mode.

Screens written against `GlassAppBar.pinned` today are already written for that future, so
step 3 is a rename-and-deprecate rather than a behavioural migration. That's also the
Flutter playbook for this kind of shift (opt-in → adopt in examples → deprecate → flip at
the major, as with `useMaterial3`). Very open to your take on whether that's the right end
state and when it should happen — it's your major to plan, and I've deliberately not
assumed a version.

## Why the chrome lives above the Navigator

Two alternatives lost on structural grounds. Counter-translating each route's own bar against
the slide needs two overlapping, partially-faded glass pills every transition, and a partial
fade of a glass surface pops its backdrop pass — the rule the modal-sheet morph documents.
A morph hand-off (hide-real/stand-in/hand-back) avoids that but has to hit a pixel-perfect
swap instant, which an interactive back-swipe starting mid-frame makes hostile. A persistent
host has neither problem, and it's literally how UIKit arranges it. The host re-provides the
scopes hoisting would otherwise lose (`DefaultButtonSettings`, `GlassIsolationScope`) and owns
its own isolated layer per the post-0.14.0 backdrop rules.

## Why the route animation is the clock, and why there's no NavigatorObserver

Every frame renders a pure interpolation between the top two registered routes'
configurations at `p = topRoute.animation.value`. Push, pop, back-swipe scrubbing and
cancelled swipes are then all the same code path — the gesture drives the same controller
Cupertino drives. Stack order comes from each route's `secondaryAnimation`, so nothing
observes navigation, which also makes it router-agnostic: any Pages-API router works through
`CupertinoApp.router(builder:)`, because its screens live in real `ModalRoute`s. I verified
this on-device with a go_router build of the same flow before dropping that demo from the
example to keep it dependency-light; the `.router` setup snippet lives in the docs guide.

One wrong turn worth recording: the registrar originally skipped offstage builds
(`TickerMode`-style guard) to keep inactive nested-navigator branches out. Flutter builds a
newly pushed route **offstage once** before the transition starts, and neither `offstage` nor
a muted ticker notifies dependents when they flip back — so every incoming route sat
unregistered for its whole transition, drawing its own bar mid-slide. The guard is gone;
ordering by `secondaryAnimation` already decides who's on top. Nested branch navigators are a
documented follow-up instead.

## Why item widths are measured

`GlassBarItem.custom` follows UIKit, which measures a `customView` during layout. A small
`RenderBox` (`_PinnedCluster`) lays out both routes' items, takes each cluster's measured
total, and interpolates its own width and every item's trailing-edge distance between the two
— no offstage measurement pass, no hardcoded per-item width, and icons just measure to the
standard 46pt slot. While hoisted, the in-route bar lays out the *real* items unpainted
(`Opacity(0)` + `IgnorePointer` + `ExcludeSemantics`) so `_ToolbarLayout`'s centred-title
constraint is byte-identical to the non-hoisted bar, custom widths included.

## Why appearing clusters switch instead of animating

When the destination has no actions (or the pop returns to the root, for the back button),
the cluster switches off/on once at the transition midpoint. I tried animating it and it
reads wrong in both directions — a scale-out drifts and shrinks, a scale-in "grows" in a way
native never does — and the directional ease compounds it: the same interpolation is walked
forwards on push and backwards on pop, so an `easeIn` exit becomes a front-loaded,
near-instant entrance when popped. Native cross-fades the glass container here, which the
no-glass-fade rule forbids, so a clean switch is the honest approximation. A *surviving*
capsule still morphs fully — that's the part that matters, and it's untouched.

## Pull-down menus, and why the chrome needed its own Overlay

Added after review: `GlassBarItem.menu` is `UIBarButtonItem.menu`, the iOS 26 overflow
button. Without it `.pinned` could not express the most common trailing item there is — both
"More" buttons in the nav-patterns demo were dead `onTap: () {}` stubs, and a screen needing
a real one had to drop back to the widget constructor and lose pinning entirely. The whole
capsule morphs into the pull-down rather than the tapped slot, matching iOS 26's
`GlassEffectContainer` and `GlassButtonGroupItem.menu`, which the in-route fallback reuses
wholesale so both paths agree.

Mostly this is wiring to `GlassMenu`, but the pinned host needs two things
`GlassButtonGroup` does not:

- **An Overlay of its own.** The chrome sits above the app's `Navigator`, which puts it
  outside the `Navigator`'s `Overlay` — and `GlassMenu` portals to the *root* overlay, so up
  there was none to find (it asserted). The shell now wraps its chrome in `Overlay.wrap`:
  full-screen and origin-aligned, so the menu's global-coordinate morph lands on its trigger
  exactly as it does inside a route, and its dismiss barrier covers the page. It's a
  *sibling* of the `Navigator`, not an ancestor, so nothing inside the app resolves its root
  overlay to this one.
- **Dismissal on navigation.** The capsule outlives every route it serves, so an open menu
  would hang over the bar while the page slid out from under it. It closes on the first
  frame the transition is no longer settled — the same flag that stops one being opened
  mid-transition.

The `GlassMenu` wrapper is applied unconditionally, even with no menu item: inserting or
removing it as an ancestor would remount the capsule's element, and a glass shell that
remounts mid-morph pops its backdrop. Closed, `GlassMenu` adds only inert wrappers and
mounts no overlay, so the empty case costs nothing.

## Back-compat

Purely additive. The plain `GlassAppBar` constructor is byte-for-byte the released API and
never participates in pinning, so existing screens are pixel-identical with or without a
shell. Participation lives entirely in the new `GlassAppBar.pinned` constructor, which has
no widget `leading`/`actions` — mixing the two APIs is structurally impossible rather than a
runtime assert. With no shell, a `GlassNavigationShell(enabled: false)`, or
`GlassQuality.minimal`, a pinned bar renders in-route as an automatic back `GlassButton`
plus a `GlassButtonGroup.icons` capsule — screens written against the new API work either
way. The title-centering regression demo (#198) deliberately keeps the widget constructor,
since asymmetric widget layout is exactly what it verifies.

Also included, found while testing: `GlassAppBar` inline titles no longer wrap to a second
line when wide actions squeeze them — single line with an ellipsis, as on iOS. That one
bites any bar with wide actions, pinned or not.

## Tests & coverage

Effective project coverage **92.61%** (gate is 90%); the four touched lib files sit at
**93.4%** combined. 51 new tests: 13 unit tests for the matching heuristics (identifier wins,
trailing-edge pairing, mismatched ids never pair, single consumption), the swap timing
(complementary directions, a surviving capsule never blinks) and the tap-handler contract,
plus 38 widget tests for registration/hand-over, the automatic back button, `onBack`, taps
at rest, intrinsic-width measurement, in-place updates, mid-transition matching, capsule-off,
chrome retreat under non-participating routes, menus (opening pinned, blocked
mid-transition, dismissed on navigation, first-item-only, in-route fallback), and every
fallback path.

The tests earned their keep twice. They caught settled bars keeping invisible zero-opacity
leftovers of the previous route's icons; and writing the mid-transition tap test surfaced a
worse one — `Route.isActive` goes false the instant a route is popped while its exit
transition still has its full duration to run, so the outgoing route dropped out of the
interpolation on frame one of a *button-driven* pop and the chrome snapped to the
destination instead of morphing into it. Back-swipes never hit it, because a swipe only pops
once the gesture commits, which is why device testing missed it. Both fixed.

`flutter analyze --fatal-warnings` clean; `dart format` clean. Verified on the iOS simulator
(Impeller): the pinned glass refracts the sliding page every frame with no ghosting, across
the showcase, the nav-patterns flows, and the nested-navigation drill-down. Rebased onto
`main` at 1.1.0 — the full non-golden suite passes (2,435 tests).

## Things to say no to if you'd rather

- **`GlassNavigationShellState.debugPinningSupported`** is a `@visibleForTesting` gate
  override — headless runs report no shader support, so without it the pinned path is
  untestable. Same pattern as `debugMorphSupportsBlending`.
- **`GlassNavPinnedMetrics` / `GlassNavActionSlot` / `matchGlassNavActions`** are public
  within `surfaces/shared/glass_nav_pinned_host.dart` but not exported from the barrel —
  reachable from tests via direct library import, the `SheetMorphGeometry` arrangement.

## Notes

- `CHANGELOG.md` entries are under `# Unreleased`, above the 1.0.0 section — move them under
  whatever version you cut next.
- New guide at `docs/GLASS_NAVIGATION_TRANSITION.md`: setup (including `.router`), matching
  rules, a behaviour table, a *Direction* section mirroring the trajectory above, and an
  explicit *Current limitations* section. `ROADMAP.md` ticks off `GlassNavigationTransition`
  and adds the "pinning becomes the default" follow-up, with the remaining parity work split
  into what is small (text styles) and what actually gates the flip (spacers / multi-capsule
  grouping / prominent items — one capsule becomes *n*, and the count can differ between the
  two routes being interpolated).
- Exports are narrow: `glass_bar_item.dart` in full, and only `GlassNavigationShell`,
  `GlassNavigationShellState`, `GlassNavBarRegistration` from the shell.
- Known follow-ups, in the docs: `spacer()` rendering (multi-capsule grouping / per-item
  `sharesBackground`), animating in-place `setState` action resizes, nested navigators
  (go_router `StatefulShellRoute` branches), RTL.


JakeThomson:
For backwards compatibility I've put the pinned behaviour behind a `GlassAppBar.pinned` constructor rather than changing the default - the widget-based leading/actions API is untouched, so existing screens render identically.

That's a transition state rather than the end goal. Longer term I think this should be the default, with the back button automatic (as Flutter's own AppBar does) and overridable when needed. I've kept that out of scope here since `GlassBarItem` can't yet express everything the widget API can (menus, spacers, text/prominent styles); once it reaches parity, a major release could flip the default and demote the widget constructor to a legacy mode. Interesting in your take on that, and the solution as a whole.

sdegenaar:
Hey @JakeThomson, again really great work. This is cool!. Good architecture and PR description is excellent. 

A few things before I can merge:

Branch has conflicts with main — please rebase before the fix-up pass, CI is blocked until then.

Three things to fix:

1. GlassBarItem.icon.onTap vs GlassBarActionItem.onTap — the factory declares it required but the base class has it nullable. Pick one and make it consistent. Either works, just needs to be intentional.
2. Back button tappable mid-transition — ignoring: !atRest only blocks when progress >= 0.999, so the button is tappable during an active swipe. Should be ignored any time the transition is moving, not just at rest. Add a widget test for this.
3. ROADMAP.md — not in the diff but you mention it ticking off GlassNavigationTransition. Include it if you don't mind.

On the trajectory — agree. Opt-in now, default at a major once GlassBarItem reaches parity.

One thing on the parity gap though — you listed menus as missing but GlassButtonGroupItem.menu already exists and is demoed in interactive_page.dart. Under GlassButtonGroup (.menu item - shared pill and group + standalone menu) Adding GlassBarItem.menu is wiring a new factory to existing infrastructure, not starting from scratch. That's a smaller gap than it sounds and might bring the "flip to default" step closer than you think.

Once those are in and the conflict resolved, mark it ready and I'll merge.

Thanks again!

Cheers

JakeThomson:
Thanks for the review! I have addressed all three callouts, and also added support for `GlassBarItem.menu` in this PR too.

Also added an example to demonstrate this and `GlassBarItem.custom`

https://github.com/user-attachments/assets/0c522513-ac88-4a02-a43f-8e7f9cdace13

PR #223, opened 2026-08-25
feat(GlassModalSheet): swipe-dismissals morph back from the release point

### LINKED ISSUE #222: [Feature]: Swipe-dismissing a morphed GlassModalSheet should morph back from the release point
### What kind of feature?

Animation or interaction improvement

Follow-up to #216. The shipped morph presentation deliberately skips the reverse morph on a swipe: the doc comment on `GlassModalSheet.show` says a drag-dismissal "keeps the sheet's own slide-away instead", because morphing back from the sheet's *resting* frame after a mid-swipe release would visibly jump.

That jump is the symptom, not the constraint. On iOS 26 (Journal, Maps, and any `.zoom`-presented sheet) a swipe below the lowest detent detaches the card from the bottom edge, shrinks it as you drag, lets you push it around on both axes, and the release morphs back into the trigger from **exactly the frame you let go at**. So a morphed sheet currently behaves natively for every dismissal except the most common one.

Swipe-dismissal joins the morph, reproducing the native gesture (numbers below are measured off iOS 26 recordings — the medium-sheet gesture was captured through iPhone Mirroring so the cursor gives the actual finger position, and fitted to an RMS scale error of 0.006):

- **Shrink normalised by the card's height, not the screen's** — one gain of ~0.64 of scale per card height dragged explains both a small panel (Maps' Map Modes, ~2.5x per screen height) and a medium sheet (~1.4x).
- **The scale pivots at the grabbed point**, carried down with the fall, so the content under the finger stays exactly under the finger. With the gain below 1.0 per card height the fall always outruns the shrink, so the card's underside never lifts into view — no bottom-pinning needed.
- **Damped travel**: 1:1 direct manipulation to ~0.48 of the card's height, then a rubber-band toward a 0.33 scale floor — a long drag parks the card near the bottom (the reference gesture parked at ~0.55 and was still creeping) instead of sliding it off while the finger is still moving. The sheet's own position keeps tracking 1:1 underneath so the dismiss threshold is unchanged.
- **The sideways axis only opens once the dismissal is under way** (`kTouchSlop` worth of travel past the detent — before that a wobble belongs to the sheet's jelly-follow stretch), anchored at the crossing so prior wander doesn't snap in.
- **Sideways is pure translation** (a 143 pt native sweep changed scale by only 0.029, all attributable to incidental vertical drift), free while there is room, and then the card's leading edge **pins at the screen edge** with only a few points of give — the largest overshoot measurable in the native sweep was ~4 pt on a 440 pt screen. Resistance follows where the card *is*, not how far the finger has moved.
- **A release short of the threshold springs back** to the detent at full size and centred, carrying the finger's release velocity; a release past it hands the exact rendered frame to the closing morph.
- **Scoped to morphed presentations** — the interactive shrink belongs to the presentation, the same split iOS draws between its zoom transition and `UISheetPresentationController` (whose detents "resize from one edge while the other three remain fixed" and cannot express this). A sheet shown without `morphFrom` keeps its plain slide-away unchanged.

None viable

### Anything else?

| Native | Current |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/0bfafa49-602f-49cc-bd4b-8d25ac91bac7" /> | <video src="https://github.com/user-attachments/assets/b47b91b9-b7c9-419a-be87-91809c3885cd" />  |

JakeThomson:
Fixes #222 — the follow-up #217 left on the table: it shipped with the doc'd behaviour that a swipe-dismissal "keeps the sheet's own slide-away", because morphing back from the *resting* frame after a mid-swipe release would visibly jump. The jump was the symptom, not the constraint — iOS 26 morphs on a swipe too, from exactly where the finger let go. Now this does.

## What changed

| Native | Before | After |
|--------|--------|--------|
| <video src="https://github.com/user-attachments/assets/0bfafa49-602f-49cc-bd4b-8d25ac91bac7" /> | <video src="https://github.com/user-attachments/assets/b47b91b9-b7c9-419a-be87-91809c3885cd" />  | <video src="https://github.com/user-attachments/assets/1f66da8f-d1f4-46b9-9f9a-480314e290c8" /> |

Below the lowest detent a morphed sheet detaches from the bottom edge, follows the finger on both axes and shrinks as it falls; the release hands the exact rendered frame to the closing morph. A release short of the dismiss threshold springs back to the detent, full size and centred, carrying its momentum. A sheet presented without `morphFrom` keeps its plain slide-away — the interactive shrink belongs to the presentation, the same split iOS draws between its zoom transition and `UISheetPresentationController`, whose detents "resize from one edge while the other three remain fixed" and so cannot express it.

No new public API: the swipe reuses the presentation #217 added. `GlassSheetMorphPresenter` gains a (library-internal) `controller` field, and `GlassModalSheetController` a private `_livePosition`, so the presenter can read the position the finger actually has rather than the one the sheet last built with.

## Where the numbers come from

The curve is measured, not tuned. The reference gesture was captured through iPhone Mirroring, where the macOS cursor is the finger — so the gain is fitted against what the hand did rather than inferred from the card's motion. Fitting gain, knee and floor jointly reproduces the measured scale at an RMS error of 0.006.

| constant | value | meaning |
|---|---|---|
| `dismissScaleGain` | 0.64 | scale lost per *card height* dragged |
| knee | 0.48 | of the card's height — 1:1 until here, rubber-banded after |
| `minDismissScale` | 0.33 | the asymptote the shrink eases toward and never reaches |

Card height, not screen height, is the normaliser: Maps' small "Map Modes" panel measures ~2.5x per screen height, a medium sheet ~1.4x, and both collapse onto one card-relative gain. Both the fall and the shrink read off a single damped travel value, so the card's size and position settle as one object — while the sheet's own position keeps tracking 1:1 underneath, leaving the dismiss threshold untouched.

## Verification

- 85 tests in the morph suite (geometry pure functions, the equivalence test, plain-sheet regression guard, unlock gating, chase behaviour, spring-back), 362 across the overlays suite, all green on top of v1.0.0.
- On-device via `example/integration_test/morph_demo_test.dart` on the iOS 27 simulator (iPhone 17, Impeller): stepped drags asserting the measured curve at every step, sideways trail bounds, spring-back, and the swipe handing off to the droplet.
- `flutter analyze` and `dart format` clean.
- Not tested on a physical device, Android, or desktop; the Skia fallback path is unchanged from #217.

## Notes

- `CHANGELOG.md` entry sits under `# Unreleased` — move it under whatever version is cut next.
- `GLASS_MODAL_SHEETS_GUIDE.md`'s "Which dismissals morph back" section is rewritten around the new behaviour, with the measurement provenance.
- Two pre-existing quirks surfaced while working on this, deliberately left alone: a large sideways move can derail the sheet's own vertical recognizer in the gesture arena, and a sheet held perfectly still shows a sub-3px transient divergence between its layout position and its live animation value. Happy to file both as issues.
- There's no analogue of `interactiveDismissShouldBegin` for vetoing the gesture from content — same story as #217, a separate issue if wanted.


sdegenaar:
Hey @JakeThomson. Great work again, this is merged — the measured curve, single-ticker horizontal chase, and dismissedRect as single source of truth are all exactly right.

A few threads if you want to follow up:

1. RenderObject-level transform — setState on every horizontal tick works but a RenderProxyBox owning the matrix directly would eliminate widget build passes mid-gesture at 120Hz. Worth a profiling run on device first.

2. Velocity-aware morph handoff — release position is frozen correctly, but the closing droplet starts from rest. Seeding the morph spring with gesture release velocity would make fast flings feel continuous.

3. Gesture arena sideways derail — you flagged this, just making sure it gets an issue so it doesn't get lost.

4. RTL — horizontalOffsetFor is left-biased. Package already has RTL consumers so screen-edge resistance will mirror on the wrong side. Directionality.of(context) flip should sort it.

None blockers — solid as landed.

Cheers

PR #227, opened 2026-08-26
feat(GlassModalSheet): seed the closing morph with the fling's velocity

JakeThomson:
Follow-up to the review comments on #223. No issue — these are the threads left on that PR.

## What changed

**Velocity-aware morph handoff.** The release position was frozen correctly, but the closing droplet started from rest, so a fast fling stalled at the point the finger left. `GlassMorphController.close()` now takes an optional `velocityHint`, and a dismissing swipe seeds it with the rate the shrink was actually running at.

The conversion falls straight out of the existing curve: the shrink runs at `dismissScaleGain` per card height dragged, and morph progress spans trigger-sized to full-sized, so px/s over card height *is* progress per second. Downward speed only, capped at 10/s, and `min`'d against the resting kick — a fling may hurry the close along, never drag it out. A barrier tap, back gesture or controller close is unchanged.

One ordering detail worth knowing: the velocity is read at the dismissal, not at the release. The sheet pops the route from its own end-of-drag handler, which runs *before* the presenter's `Listener` sees the pointer up — so `_onRouteStatus` fires first and a value stored on release would always be a gesture late.

**RTL — no change, and I'd argue against making one.** `horizontalOffsetFor` has no leading/trailing concept: slack is `cardRect.left` going left and `screenWidth - cardRect.right` going right, and the sheet's own layout is symmetric (a single `horizontalMargin`, a centred `peekWidth`, non-directional `Positioned`). It already mirrors. A `Directionality.of(context)` flip would move the screen-edge resistance to the *wrong* side for RTL users, which is the bug the thread was trying to prevent. Pinned both ways instead — a mirror-symmetry unit test, and a widget test running the identical sweep under `TextDirection.rtl` and `.ltr` and asserting they land within 0.5px.

## The RenderObject transform, measured and dropped

The suggestion was that `setState` per horizontal tick costs a build pass over the sheet subtree at 120Hz. It doesn't: `widget.child` is a field on the presenter, not something its `build()` recreates, and `Element.updateChild` returns early at an identical widget instance. The rebuild stops dead at the sheet.

Measured on the pre-change code — the card moves and rescales while the sheet's content builds zero times:

```
counter works, builds so far=1              ← the counter does increment
moved=112.05  scaleChanged=true  builds=0   ← during the chase
```

I built the `RenderProxyBox` version anyway and profiled it on a physical iPhone (iOS 27, profile mode, 270 frames per run, 5 drag passes, first pass discarded as warm-up):

| | build mean | build p90 | build p99 | raster mean | frames > 8.3ms |
|---|---|---|---|---|---|
| RenderProxyBox | 0.37 ms | 0.98 | 2.09 | 2.84 | 0 / 270 |
| `Transform` + setState | 0.41 ms | 1.07 | 2.18 | 2.99 | 0 / 270 |

Parity, and the difference is well under the run-to-run noise (repeat runs varied raster by ~1.5 ms). So it bought ~100 lines re-implementing `RenderTransform`, plus a real footgun — the matrix has to be recomputed at the end of `onPointerMove` because of the handler ordering above, and getting that wrong slides the shrink pivot off the finger. Not worth it; reverted.

Happy to revisit if someone has a sheet heavy enough to change the answer.

## Verification

- 94 tests in the morph suite, 416 across the overlays and morph-controller suites, green on top of v1.1.0.
- Physical device (iPhone, iOS 27) for the profiling above; the fling handoff itself was checked on the iOS 27 simulator.
- `flutter analyze` and `dart format` clean.
- Not tested on Android or desktop; the Skia fallback path is untouched.

## Notes

- `CHANGELOG.md` entry sits under `# Unreleased` — move it under whatever version is cut next.


sdegenaar:
Thanks Jake — again great work! Applied this directly to main alongside the review follow-ups from #221 so everything lands in one clean commit before the 1.1.0 tag.  Closing rather than merging to keep the history tidy.

Much appreciated,
Cheers

PR #236, opened 2026-08-27
refactor(nav): expose the pinned-chrome registrar as GlassPinnedBarChrome

### LINKED ISSUE #235: [Feature]: A public registrar so bars that aren't GlassAppBar can pin
### What kind of feature?

API / developer experience

Pinning is currently reachable only through `GlassAppBar.pinned`. An app whose bars are its own widgets — a Material `AppBar` with a bespoke backdrop, a collapsing large-title sliver — cannot opt into a `GlassNavigationShell` without reimplementing the registration handshake.

The public API is already sufficient: `GlassNavigationShell.maybeOf`, `GlassNavigationShellState.register` / `unregister`, and `GlassNavBarRegistration` are all exported. What's missing is the widget that uses them correctly. The one that exists is private — `_GlassNavBarRegistrar` in `lib/widgets/surfaces/glass_app_bar.dart`.

So every consumer in that position rewrites ~60 lines, including three subtleties that are not obvious from the outside and easy to get wrong:

1. **The deferred hand-over.** The bar must keep drawing its own chrome until the shell has had a frame to render its copy — hence the `addPostFrameCallback` before flipping to hoisted. Get it wrong and the two either overlap for a frame or both vanish.
2. **Release on change, not just on dispose.** The registration must be dropped and re-taken when either the shell or the `ModalRoute` changes identity.
3. **No offstage / `TickerMode` guard.** Flutter builds a newly pushed route offstage once before the transition starts, and neither flag notifies dependents when it flips back — so skipping those builds strands the route unregistered for the whole transition. The existing private registrar has a comment explaining exactly this; a reimplementer will not know it.

This is not hypothetical: an app adopting `GlassNavigationTransition` incrementally alongside an existing design system wrote a ~60-line reimplementation against the public API precisely because the registrar is private. That seems likely to be the common case for any app with bars it already owns.

Promote the registrar to public API, and refactor `GlassAppBar.pinned` to use it — so there is one implementation rather than two. Sketch:

```dart
GlassPinnedBarChrome(
  actions: [
    GlassBarItem.icon(icon: const Icon(CupertinoIcons.add), onTap: _create),
  ],
  builder: (context, hoisted) => AppBar(
    title: const Text('Repository'),
    automaticallyImplyLeading: !hoisted,
    leading: hoisted ? const SizedBox(width: 44) : null,
    actions: hoisted ? null : [IconButton(icon: addIcon, onPressed: _create)],
  ),
);
```

`backButton`, `onBack` and `buttonSettings` would mean what they do on `GlassAppBar.pinned`.

Design points I'd want your call on:

- **The name.** `GlassPinnedBarChrome` should read as "hand my chrome to the shell", not as a bar. Happy to change it.
- **Whether to refuse non-dismissible routes outright.** The consuming app guards on `route.impliesAppBarDismissal`, because a root route has nothing to pin *and*, in an app with a nested navigator (a go_router `ShellRoute` for tabs), registering from the nested stack ranks routes from two different navigators against each other — which the shell's ordering cannot express. Building it in would make the one-navigator-per-shell rule self-enforcing. But `GlassAppBar.pinned` deliberately *does* pin a root route's actions today (and is tested for it), so making it unconditional would be a behaviour change. My inclination is an `enabled` flag the consumer drives, documented with the nested-navigator recipe.
- **Whether `buttonSettings` belongs on this widget** or stays a `GlassAppBar` concern.

Reimplementing the handshake against the exported `register`/`unregister` API. It works, but it's a copy of code the package already has, and the three subtleties above are only discoverable by reading the package source.

### Anything else?

Happy to open the PR — the diff is probably the clearest form of this proposal, and it's a refactor rather than an addition once `GlassAppBar.pinned` routes through the same widget.

JakeThomson:
## Description

Pinning is currently reachable only through `GlassAppBar.pinned`. An app whose bars are its own widgets — a Material `AppBar` with a bespoke backdrop, a collapsing large-title sliver — cannot opt into a `GlassNavigationShell` without reimplementing the registration handshake, because the widget that performs it (`_GlassNavBarRegistrar`, `lib/widgets/surfaces/glass_app_bar.dart`) is private.

Everything else it needs is already public: `GlassNavigationShell.maybeOf`, `GlassNavigationShellState.register` / `unregister`, and `GlassNavBarRegistration` are all exported. So a consumer in that position rewrites ~60 lines, including three subtleties that are only discoverable by reading the package source:

1. **The deferred hand-over.** The bar must keep drawing its own chrome until the shell has had a frame to render its copy — hence the `addPostFrameCallback` before flipping to hoisted. Get it wrong and the two either overlap for a frame or both vanish.
2. **Release on change, not just on dispose.** The registration must be dropped and re-taken when either the shell or the `ModalRoute` changes identity.
3. **No offstage / `TickerMode` guard.** Flutter builds a newly pushed route offstage once before the transition starts, and neither flag notifies dependents when it flips back — so skipping those builds strands the route unregistered for the whole transition. The private registrar has a comment explaining exactly this; a reimplementer will not know it.

This PR promotes that widget to `GlassPinnedBarChrome` and **refactors `GlassAppBar.pinned` to use it**, so there is one implementation rather than two. No rendered pixel changes and every existing `GlassAppBar.pinned` test passes unchanged.

```dart
GlassPinnedBarChrome(
  actions: [
    GlassBarItem.icon(icon: const Icon(CupertinoIcons.add), onTap: _create),
  ],
  builder: (context, chrome) => AppBar(
    automaticallyImplyLeading: false,
    leading: chrome.leading,
    actions: chrome.actions,
    title: const Text('Repository'),
  ),
);
```

Items are declared **once**, as data. `chrome.leading` and `chrome.actions` hold the real glass back button and actions capsule until the shell has taken them, and same-sized unpainted placeholders after — so the bar never builds a second copy of anything, and there is nothing to keep in sync with the item data. `chrome.hoisted` is there for a bar that wants to substitute its *own* chrome, but the common case never reads it.

That construction is exactly what `GlassAppBar._buildBar` was doing inline, and it is a pure function of the item data, so moving it into the shared widget takes ~60 lines of branching out of `GlassAppBar` too. The two paths are genuinely one implementation now rather than one wrapper plus two copies of the rendering.

Fixes #235

### Design points — happy to change any of these

- **The name.** `GlassPinnedBarChrome` is meant to read as "hand my chrome to the shell", not as a bar. Say the word if you'd rather it were something else.
- **`enabled` rather than a built-in `impliesAppBarDismissal` guard.** A guard on non-dismissible routes would make the one-navigator-per-shell rule self-enforcing, which is tempting — an app with a nested navigator (a go_router `ShellRoute` for tabs) otherwise ranks routes from two different navigators against each other, which the shell's ordering cannot express. But `GlassAppBar.pinned` deliberately *does* pin a root route's actions today, and is tested for it, so building the guard in would be a behaviour change rather than a refactor. Instead there's an `enabled` flag the consumer drives, documented with the recipe:
  ```dart
  enabled: ModalRoute.of(context)?.impliesAppBarDismissal ?? false,
  ```
  Happy to make it the default instead if you'd rather take the behaviour change.
- **`buttonSettings` stays on this widget**, since it is part of `GlassNavBarRegistration` and a non-`GlassAppBar` bar has no other way to set it. It is applied to the in-route buttons as well, via `DefaultButtonSettings`, so the two look identical across the hand-over.
- The `GlassBarItem.spacer()` assert moved from `GlassAppBar.build` to `GlassPinnedBarChrome.build`. It still covers both the pinned and in-route fallback paths — the chrome widget wraps both — and now guards the new public entry point too.

### Where this is coming from

An app adopting `GlassNavigationTransition` alongside an existing design system hit this and wrote a ~60-line reimplementation against the public API. Its bars are Material `AppBar`s carrying their own progressive-blur backdrop and collapsing large titles, not `GlassAppBar`s. That seems likely to be the common shape for any app pinning incrementally.

There is a sibling one-liner I'd like to send separately — exporting `GlassNavPinnedMetrics`. It is less pressing now that the slots come pre-sized, but it is still the only way to reason about the pinned cluster's geometry from outside. Same story, *support pinning from a bar that isn't `GlassAppBar`*, and independent of the API review here.

## Type of Change

- [x] 🧹 Refactor / cleanup (no behavior change)

## Platforms Tested

- [x] iOS (Impeller)
- [ ] Android (Impeller)
- [ ] macOS
- [ ] Windows
- [ ] Web
- [ ] Linux

Verified on an iPhone 17 simulator running iOS 26.0 — glass is shader-drawn there, so no device was needed. Not run on any other platform; nothing in the diff is platform-specific (it is registration bookkeeping, not rendering).

## Checklist

- [ ] If this changes shader code, I have verified it compiles on both Metal (macOS) and SkSL/SPIR-V (Windows) — n/a
- [x] No dynamic indexing or unsupported GLSL builtins were introduced in shader code — n/a

**On tests:** all 2716 non-golden tests pass. 11 golden tests fail, but they fail identically on `main` at `c93e7de` with no changes applied — they look environment-sensitive here rather than related to this diff, and nothing in it touches those widgets. I deliberately did not add or regenerate goldens: glass is shader-drawn and environment-sensitive, so a regenerated golden would prove nothing.

New widget tests in `test/widgets/surfaces/glass_pinned_bar_chrome_test.dart` cover: registering on mount from a plain Material `AppBar`; `hoisted` flipping only on the frame *after* registration; unregistering when the bar leaves the tree; a pushed route taking the chrome and handing it back on pop; `onBack` and `backButton: false`; the in-route back button and capsule being built for the consumer and working with no shell at all; unsupported device / `enabled: false` falling back in-route; and flipping `enabled` off releasing a live registration. They use `GlassNavigationShellState.debugPinningSupported`.

## Screenshots / Recordings

A new **Custom Bar Pinning** pattern in `nav_bar_patterns_demo.dart` — a two-level drill-down whose bar is a plain Flutter `AppBar`, which is the clearest review artifact for this PR:

```bash
cd example
flutter run -t lib/demos/nav_bar_patterns_demo.dart   # → "Custom Bar Pinning"
```

Verified on iOS 26.0: at the root the `AppBar` shows its title with no leading and the glass capsule (`+`, search) sits pinned at the trailing edge; mid-push the capsule holds exactly still while the `AppBar` and its content slide out from under it; on the pushed route the pinned back button has appeared and search has cross-faded into `…` with `+` holding its position (same `id`).\n\nThe demo screen is ~90 lines and contains no `hoisted` branch at all.


JakeThomson:
Reshaped the API in c0a7063 after a review point that I think was the right call.

The first cut handed the builder a `hoisted` bool and stopped there, which meant a consumer declared every action **three** times — once as data, once as an in-route fallback widget, once as a same-sized placeholder — and had to keep all three in sync, including hardcoding the back button's 44pt diameter to match a constant the package owns:

```dart
// before — three declarations of one action
actions: [GlassBarItem.icon(icon: const Icon(CupertinoIcons.add), onTap: _create)],
builder: (context, hoisted) => AppBar(
  leading: hoisted ? const SizedBox(width: 44) : null,
  actions: hoisted ? null : [IconButton(icon: addIcon, onPressed: _create)],
),
```

That was strictly worse than the private widget it came from: `GlassAppBar.pinned` asks for the data once and builds both the in-route capsule and the measuring placeholder itself. I had extracted the registration half and left the rendering half behind.

Since that rendering is a pure function of the item data — nothing in it is `GlassAppBar`-specific — it moves into the shared widget, and the builder now receives finished slots:

```dart
// after — one declaration
actions: [GlassBarItem.icon(icon: const Icon(CupertinoIcons.add), onTap: _create)],
builder: (context, chrome) => AppBar(
  automaticallyImplyLeading: false,
  leading: chrome.leading,
  actions: chrome.actions,
),
```

`chrome.hoisted` is still there for a bar that wants to substitute its own chrome, but the common case never reads it. Two side effects worth noting:

- It takes ~60 lines of branching out of `GlassAppBar._buildBar` as well, so the two paths are now one implementation rather than one wrapper plus two copies of the rendering. Net `glass_app_bar.dart` is −221/+14.
- The first frame is now genuinely invisible: before hoisting the bar draws the *same* `GlassButtonGroup.icons` capsule the shell is about to draw, rather than the consumer's own approximation of it.

The example screen lost its `hoisted` branch entirely. Still no behaviour change to `GlassAppBar.pinned` — its existing tests pass untouched.

PR #238, opened 2026-08-27
feat(GlassAppBar): add a leading API to pinned navigation chrome

### LINKED ISSUE #237: [Feature]: a leading API for GlassAppBar.pinned
### What kind of feature?

New property or option for an existing widget

`GlassAppBar.pinned` can express exactly one leading control: its own automatic back button. `GlassNavBarRegistration` has no way to say anything else about the leading side, so a screen with a custom leading — a Cancel, a close X, a stage-dependent button, a profile photo — cannot use pinned chrome at all.

Half-pinning is not an option either. A screen could pin only its trailing cluster, but that leaves the leading capsule sliding with the page while the trailing one stays put, which looks worse than not pinning. Falling back whole is the honest result, so those screens lose the pinned behaviour entirely.

In one app that has just switched its shared app bar to pin by default on iOS 26+, of 136 call sites:

- ~18 pass the standard back button (special-cased as such, and pin fine)
- ~79 pass no leading at all (pin fine)
- **~39 pass a genuinely custom leading and are excluded from pinning** — 28 bare `IconButton`s, 3 close buttons, 2 `IconButton(Icons.close)`, 1 `BackButton`, and ~5 conditional or `BlocBuilder`-wrapped leadings

So the gap costs roughly 29% of that app's bars.

`leading` on `GlassAppBar.pinned`, taking the same `GlassBarItem` vocabulary as `actions`:

```dart
GlassAppBar.pinned(
  title: const Text('New Issue'),
  leading: [
    GlassBarItem.icon(
      icon: const Icon(CupertinoIcons.xmark),
      label: 'Cancel',
      background: GlassBarItemBackground.separate,
      onTap: () => Navigator.of(context).pop(),
    ),
  ],
)
```

Two behaviours iOS 26 has that the current API cannot express:

**A custom leading, replacing or supplementing the back button.** `UINavigationItem.leftBarButtonItems` is plural, and `UINavigationItem.h` documents the default: *"A custom left item replaces the regular back button unless you set leftItemsSupplementBackButton to YES."* Flutter's own `AppBar` reaches the same rule from the other direction — it implies a leading only when `leading == null`. So `leading` should replace, with a `leadingItemsSupplementBackButton` flag mirroring the UIKit property for the case that wants both.

**A leading that carries no glass.** iOS 26 added two booleans to `UIBarButtonItem`: `hidesSharedBackground` (*"prevent the standard shared background (typically using the Glass effect) from being drawn behind this bar button item"*) and `sharesBackground` (its own background, grouped with nothing). A profile photo in the leading slot is the first of those — it is already a circle, and a capsule behind it reads as a second, competing surface. Their four combinations describe three results, which suggests an enum on the item rather than two interacting bools.

The consuming app excludes any bar with a non-standard leading from pinning, with a `_canPin` getter that only accepts its own `AdaptiveBackButton`. It works, but it is what costs the 29% above — those screens keep the in-route fallback while the rest of the app pins.

Pinning only the trailing cluster was considered and rejected: the leading capsule would slide with the page while the trailing one stayed put.

### Anything else?

**Feasibility.** The leading side looks like a degenerate case of machinery the trailing side already has. `_PinnedCluster` / `_RenderPinnedCluster` already does intrinsic-width measurement, width interpolation, `id` matching and icon cross-fade; what it lacks is a direction. It records every slot's distance from the cluster's *trailing* edge, and a leading cluster has to anchor to the leading edge and grow trailing-ward.

Two things worth checking before committing to that read:

- **A lone back button must stay pixel-identical.** Today it is one 44pt circular `GlassButton`, and that circle *is* the iOS 26 look — it is not a one-item capsule.
- **Relationship to the unshipped spacer work.** A leading cluster is effectively a second capsule, and the roadmap already lists `GlassBarItem.spacer()` / multi-capsule grouping as structural work on the same render object.

**Adjacent, worth fixing in the same place.** `_PinnedBackButton` passes `label: 'Back'` as an English literal, so the pinned back button's accessibility label is untranslated in every non-English app. Pre-existing rather than caused by this, but it is the exact widget this work touches.

Implemented in #238.

JakeThomson:
Closes #237.

Adds `leading` to `GlassAppBar.pinned`, so a screen with a custom leading control can use pinned chrome instead of falling back whole.

```dart
GlassAppBar.pinned(
  title: const Text('New Issue'),
  leading: [
    GlassBarItem.icon(
      icon: const Icon(CupertinoIcons.xmark),
      label: 'Cancel',
      background: GlassBarItemBackground.separate,
      onTap: () => Navigator.of(context).pop(),
    ),
  ],
)
```

https://github.com/user-attachments/assets/b0706dfb-b2eb-4cb2-9377-fda369d8321a

(the flashing seen when navigating is fixed in https://github.com/sdegenaar/liquid_glass_widgets/pull/240)

## Feasibility read

The issue's guess was right: this is a generalisation of `_PinnedCluster`, not new machinery, and the render object's anchoring is the only assumption that had to be lifted.

`_RenderPinnedCluster.performLayout` already walked leading-to-trailing accumulating `consumed`, then *derived* the trailing distance as `total - consumed - w`. Anchoring to the leading edge is the same walk reading `consumed` directly. It now stores distance-from-the-anchored-edge and places children at `x = anchoredAtStart ? d : size.width - d - slotWidth`. `computeDryLayout` needed no change — it only sums totals.

Everything else that looked directional was a parameter: `matchGlassNavActions` takes the anchor (positional slots count from the edge the cluster grows away from), the host's `Positioned(left/right: 0)` became `Positioned.directional`, and the coverage-retreat `Transform.scale` alignments were already correct per side.

`_PinnedActionsCapsule` became `_PinnedGroup`, which renders any shell on either side. Leading clusters therefore inherit width interpolation, `id` matching, icon cross-fade, tap-swallowing while `!settled`, menu dismissal when navigation starts, and the coverage retreat scaled about each cluster's own anchored edge — none of it reimplemented.

## API

Three knobs, mapping onto UIKit's three:

| | UIKit |
|---|---|
| `leading` | `leftBarButtonItems` |
| `leadingItemsSupplementBackButton` | `leftItemsSupplementBackButton` |
| `backButton` | `!hidesBackButton` |

A non-empty `leading` **replaces** the back button — `UINavigationItem.h`: *"A custom left item replaces the regular back button unless you set leftItemsSupplementBackButton to YES"* — which is also Flutter's `AppBar` rule, since it implies a leading only when none was given. `backButton: false` still wins over either. `onBack` stays orthogonal, and remains the analogue of `UINavigationItem.backAction`.

`GlassBarItemBackground` is new on every item kind, defaulting to `shared`:

```dart
enum GlassBarItemBackground { shared, separate, none }
```

iOS 26 added *two* booleans to `UIBarButtonItem` and the header distinguishes them: `sharesBackground` (default `YES` — joins its neighbours' capsule; `NO` gives it its own) and `hidesSharedBackground` (*"prevent the standard shared background (typically using the Glass effect) from being drawn behind this bar button item"*). Their four combinations describe three results, which are the three values here. `none` is what lets a profile photo sit in the bar without a capsule competing with it.

I considered dropping `leadingItemsSupplementBackButton` for a composable `GlassBarItem.back()`, and rejected it: UIKit has no bar-button factory for the back button, and neither does Flutter's `AppBar`. It read better but would have been inventing vocabulary where Apple already has some.

## A lone back button is unchanged, with no special case

`LiquidOval.effectiveRadius` is `double.infinity`, clamped in `glass_effect.dart` to `min(w, h) / 2`. On a 44×44 box that is 22 — the same shader corner radius `LiquidRoundedRectangle(22)` produces — and `getOuterPath` on a 44×44 RRect with radius 22 is a circle. So one shape draws both the circle and the capsule, and only the box has to differ.

That falls out of `GlassBarItemBackground` rather than needing a rule of its own: the automatic back button is a `separate` item, and a `separate` group lays out at `backDiameter` (44) with `GlassButton`'s default stretch, where a `shared` group keeps `slot` (46) and `capsuleStretch`. No leading-versus-trailing asymmetry and no item-count heuristic — the geometry is a property of how an item shares its background, which is what Apple's two booleans encode.

**The existing 709 surface tests pass unmodified**, and a new test pins the geometry explicitly:

```dart
expect(tester.getSize(inHost(find.byType(GlassButton))), const Size(44, 44));
```

The tempting shortcut — keep the old `_PinnedBackButton` when `leading` is empty, and only build a cluster otherwise — is wrong: pushing from a back-only route to a `leading: [Cancel]` route would swap widget types at the leading position, remounting the glass element mid-morph, and a glass shell that remounts pops its backdrop.

## Localised back label

`_PinnedBackButton` passed `label: 'Back'` as an English literal, so the pinned back button's accessibility label was untranslated in every non-English app. It now reads `CupertinoLocalizations.backButtonLabel`, falling back to the same string for apps shipping no delegates — no new dependency, no ARB pipeline. These were the only two user-visible string literals in `lib/`.

## Tests

15 new widget tests in `glass_nav_pinned_leading_test.dart`, no goldens — glass is shader-drawn and a regenerated golden would prove nothing:

- back-button geometry (44×44), localised label, and the no-delegate fallback
- replace, supplement, `backButton: false` winning over both, and a leading on a root route
- a `none` item drawing no shell; a `separate` item as its own 44pt shell beside a 46pt shared capsule
- leading-edge anchoring — a matched item holds its `dx` while the cluster grows trailing-ward, which is what would regress if the render object stayed trailing-anchored
- `matchGlassNavActions` positional matching in both directions
- leading items tappable at rest, inert mid-transition
- in-route fallback for both glass and bare leadings
- the spacer assert now covering `leading` too

Full suite: 2731 pass. The 11 golden failures on this machine fail identically on an unmodified checkout — environment-sensitive, not from this change.

## Demo

`example/lib/demos/nav_bar_patterns_demo.dart` gains a **Leading Items** pattern that walks four configurations: a bare avatar (`background: none`), the implied back button, a leading that replaces it, and one beside it. The trailing capsule holds its `id`-matched items throughout, so the leading side is the only thing moving. Verified on an iOS 26 simulator.

```bash
cd example
flutter run -t lib/demos/nav_bar_patterns_demo.dart
```

## Deliberately out of scope

Recorded in the feature doc's limitations rather than half-done:

- **`GlassBarItem.spacer()` still asserts.** The group list it needs now exists — a spacer becomes a run split — but matching *groups* across routes when the counts differ is the hard half and wants its own PR. Nothing here has to be undone for it.
- **A lone `shared` item still renders at 46**, so it is 2pt larger than a lone `separate` one. Unifying them would resize every shipped one-action bar, so it is left for a major.
- **Two wide clusters can still meet in the middle.** iOS 27 answers this with `UIBarButtonItem.visibilityPriority` and `ToolbarOverflowMenu`; inventing a clamp now would only have to be unpicked.
- **RTL.** Clusters are now placed with `Positioned.directional` and anchored to the logical edge, which is strictly better than the previous hardcoded `left`/`right`, but item order within a cluster is not mirrored. Still untested, as before.

Additive throughout — every new parameter has a default that reproduces current behaviour, and `GlassNavBarRegistration` grows one defaulted field, so existing consumers of the public shell API compile and behave unchanged.

Checked against both the iPhoneOS26.0 and iPhoneOS27.0 SDK headers; nothing this mirrors changed between them.


JakeThomson:
Follow-up pushed in 694effd.

While reviewing demo coverage I found a code path in this PR with neither a test nor a demo. A group that shares with nothing lays out at the 44pt back-button diameter, a shared capsule at the 46pt icon-slot height — so pushing from a lone leading to a two-item shared one interpolates the shell's **height** as well as its width:

```dart
height: lerpDouble(fromGroup.height, toGroup.height, eased)!,
```

Nothing exercised it. The anchoring test goes shared → shared, where both sides are 46pt and nothing moves vertically, and the demo's three transitions were bare→glass, separate→separate, and a group entering beside one that holds. A 2pt height change mid-morph is exactly the kind of thing that reads as a wobble, so it deserved better than being inferred from the code.

Now covered by a test asserting both dimensions strictly between the two shells at 150ms in and `Size(92, 46)` at rest, and by a **Circle Grows Into Capsule** stage in the Leading Items demo. Verified on an iOS 26 simulator.

Also filed the two deferred items as their own issues rather than leaving them buried in this PR's description:

- #244 — two wide clusters can overlap, with no clamp and no assert. Pre-existing (a wide custom trailing item could already collide with the back button), but `leading` widens the exposure. Suggests an assert now and `visibilityPriority` when iOS 27's shape settles.
- #245 — what is actually left for `GlassBarItem.spacer()` now that the group model landed here. The notable one: `_PinnedSide` pairs groups **by index**, which is fine while the count is stable but wrong once a spacer makes it vary between routes.

And the roadmap now records the 2.0 behaviour change this PR deliberately did not make: a group holding a single `shared` item still renders 2pt larger than the back button beside it, and unifying them would resize every shipped one-action pinned bar.

sdegenaar:
Hey @JakeThomson — fantastic work across the board, a huge amount landed today. #240 is merged and looking great.

Two PRs left before we cut 1.2.0:

#238 (your leading API) — just needs the conflict in glass_nav_pinned_host.dart resolved. Since you authored both sides (#238 and #240), you're the best person to combine them. The rest auto-resolved.

#243 (Marco's gel morph) — same conflict situation. Marco did solid work here and I'd love to get it in too. Your materialize (#240) is beautifully smooth and Marco's gel morph is technically impressive — ideally they'd both ship and compose together rather than being competing choices. Thoughts here? Could you review #243 and help unblock it? Your knowledge of the pinned host makes you the right set of eyes.

Really appreciate it — this is shaping up to be a strong 1.2.0.

Cheers

JakeThomson:
> Hey @JakeThomson — fantastic work across the board, a huge amount landed today. #240 is merged and looking great.
> 
> Two PRs left before we cut 1.2.0:
> 
> #238 (your leading API) — just needs the conflict in glass_nav_pinned_host.dart resolved. Since you authored both sides (#238 and #240), you're the best person to combine them. The rest auto-resolved.
> 
> #243 (Marco's gel morph) — same conflict situation. Marco did solid work here and I'd love to get it in too. Your materialize (#240) is beautifully smooth and Marco's gel morph is technically impressive — ideally they'd both ship and compose together rather than being competing choices. Thoughts here? Could you review #243 and help unblock it? Your knowledge of the pinned host makes you the right set of eyes.
> 
> Really appreciate it — this is shaping up to be a strong 1.2.0.
> 
> Cheers

For sure! I starting to feel slightly bad for invading the project over the last week 😄 we're doing a huge refactor of our app to liquid glass, and these are just items that have popped up along the way. 

Those two animation PRs definitely do go hand in hand - happy to take on the review

sdegenaar:
Thanks @JakeThomson. Thats exciting you moving to liquid glass. Love to see screens when ready. No keep it all coming. Really appreciate the contribution and help!! Its all high quality contributions. 

Cheers

PR #243, opened 2026-08-27
feat(GlassNavigationShell): native gel morph for the pinned actions capsule

### LINKED ISSUE #242: GlassNavigationShell: pinned capsule transitions read as a resize, not the native iOS 26 gel morph
#221 landed the pinned chrome model: the back button and actions capsule stay put while pages slide, and a surviving capsule morphs in place. The *choreography* of that morph doesn't match what the native bar actually does, though. Comparing side-by-side against a SwiftUI `NavigationStack` running the same screens (real iOS 26 bar, same simulator):

**Native:**
- The whole pill **inflates** as the morph starts — width, height, corner radius and glyphs together, peaking around +20% — then springs to the target cluster's size, overshooting and settling back. Measured frame-by-frame on a 2-icon → 2-icon push: rest height 44pt → ~53pt peak at about a third of the transition, ~2% squeeze past final size, settle.
- The spring settles **in the same breath the page lands** — capsule and page share one clock.
- Glyphs never plainly cross-fade: an outgoing glyph smears away under heavy blur while the pill reshapes beneath it; an incoming one arrives soft and sharpens last.
- A pop plays the same forward choreography toward the other cluster — not the push in reverse.
- A cluster that exists on only one side gels in/out through scale rather than appearing at an instant.

**Current:** capsule width glides between cluster widths on an ease, glyphs opacity-cross-fade in place, appearing/disappearing clusters hard-switch at the transition midpoint, and a pop is the push played backwards.

Happy to contribute this — I have it working, measured against the native reference, with the choreography pinned by a frame-by-frame trace test. PR incoming.

marco242424:
Fixes #242 — the pinned capsule now reshapes the way the iOS 26 bar does.

| Native (SwiftUI) | Before | After |
|---|---|---|
| <video src="https://github.com/user-attachments/assets/b66c2f85-4484-4ff8-a2be-ec4f9484269f" /> | <video src="https://github.com/user-attachments/assets/b87fb6ca-aea3-49ad-b71f-74a6e220bb6d" /> | <video src="https://github.com/user-attachments/assets/1a53d20b-c6e6-4ee5-a21c-5d248ff22921" /> |

All three videos are the same drill-down: 2→1 icons, an identifier-matched 2→2 swap, and 2→3, with pops. The native column is a real `NavigationStack` with `ToolbarItem`s, recorded on the same simulator.

## What changed

One code path — the pinned host. Nothing outside `GlassAppBar.pinned` chrome is touched, and the public API is unchanged.

**The capsule and the page share one clock.** Cluster width and item positions ride `GlassSpring.bouncy` — the same profile the liquid morphs elsewhere in the package use — sampled as a curve over route progress, from the first frame of the transition to the last, settling in the same breath the page does. The spring's overshoot is deliberately not walked in width: it is expressed as the whole-shell squeeze, so width travel stays monotone while the bounce reads on the entire pill. Everything stays a pure interpolation of the route clock (the invariant from #221 is preserved).

**The gel is real geometry.** The swell pulse and landing squeeze scale the pill's actual layout — the cluster lays out at scale, the glass shell re-renders its true shape, the corner radius scales, and paint carries the glyphs with it. Deliberately *not* a paint transform: `useOwnLayer` glass gets ~12px of texture headroom (sized for the ~5% press scale), and a +22% transform clips the shell at its texture edge while the contents keep painting — the shell/content disconnect this replaces. Constants were calibrated against frame-by-frame measurements of the native bar (swell +22% peaking at ~34%, ~3% squeeze) and live together in `GlassNavPinnedMetrics`.

**Glyphs smear, they don't fade.** An outgoing glyph blurs ahead of its fade and is gone by mid-morph; an incoming one arrives soft (σ3.5) and sharpens last, just before the landing. Item contents aren't glass, so filtering them is safe — the shell itself still never animates opacity.

**A pop is a forward morph.** The host mirrors the clock and swaps the from/to roles, so the swell leads and the bounce lands with the page in both directions. Rendering the push in reverse put all the motion at the wrong end of a pop. A cancelled swipe walks the mirrored clock back to its own resting frame.

**Appearing and disappearing clusters gel through scale.** This replaces the deliberate hard-switch at `swapAt` — measured natively, an emptied cluster scales out and a new one scales in, and scale is the one dimension glass can safely animate. The back button gets the same treatment on the first push off the root and the pop back to it.

**Two fixes included:** a one-frame flash of the destination chrome at push start (a just-pushed route's proxy animation reports completed-at-1.0 for its first frame; the shell now pins that frame to progress 0), and blurred glyphs drifting off the pill mid-morph (composited blur layers ignore canvas transforms; the gel scale now rides a transform layer).

## Testing

`glass_nav_morph_test.dart` pumps real pushes and pops frame-by-frame and asserts the choreography, not just endpoints: monotone width travel under the gel, swell-before-squeeze ordering, exact landing at scale 1.0, the pop leading with its swell, and the back button's bounce past full size. `--dart-define=GLASS_NAV_MORPH_TRACE=true` prints the per-frame trace — it's the harness these constants were tuned with, against the native recording. Presence tests updated for the gel semantics. 2,484 tests green under the CI selection.

The example's Navigation Patterns page gains **Header Actions Morph** — a repository-style drill-down where each row pushes a different trailing cluster (contract to one, identifier-matched hold, widen to three), each scrubbed by the back-swipe.


sdegenaar:
Hey @marco242424 — great work on the gel morph, we really want this in 1.2.0. A lot has landed on main since you raised this (materialize transitions, pinned leading API, GlassPinnedBarChrome refactor) which is why you're seeing conflicts in glass_nav_pinned_host.dart.

@JakeThomson is reviewing your PR now and knows the file well — he'll help clarify what needs to resolve. Once the rebase is done we can merge immediately.

Thanks for your patience, this one's worth getting right. Thanks again and great work!!!!!

Cheers

marco242424:
Hi @sdegenaar,

No worries i've done it. Should be all working now!

JakeThomson:
Few items to call out:

## 1. The gel fires when nothing has changed 🔴

```dart
final morphing = fromItems.isNotEmpty && toItems.isNotEmpty;
```

There's no check that the two clusters actually *differ*. I pushed B→C where both routes carry identical actions and an identical back button:

| | rest | peak | rest |
|---|---|---|---|
| trailing capsule | 92×46 | **112×56** | 92×46 |
| back button | 44×44 | **54×54** | 44×44 |

So the commonest navigation in any app — list → detail with an unchanged toolbar — now makes both the capsule and the back button balloon 22% and squeeze back for no reason. Natively an unchanged toolbar sits still through a drill-down; before this PR the case was visually static (equal widths through the `easeInOut` lerp).

Could the gel be gated on the morph having actual work to do — slot count differing, or any matched item cross-fading? Everything else in the choreography can stay as-is.

## 2. `_RenderPinnedCluster` transforms in `paint()` without `applyPaintTransform` 🔴

`paint()` pushes `Matrix4.diagonal3Values(morphScale, morphScale, 1)`, but `hitTestChildren` is a bare `defaultHitTestChildren` and there's no `applyPaintTransform` override. Measured mid-morph: the cluster box reports 55.6pt tall (scale 1.208) while its child icons still report 24×24 at unscaled offsets.

So for the whole transition, hit regions, semantics rects and descendant `localToGlobal` disagree with what's painted. Items are `enabled: state.settled` so taps mostly don't bite, but a `GlassMenu` open across a push anchors its overlay wrong and a11y focus rects are misplaced. Standard rule for a custom `RenderBox`: mirror the paint transform in `applyPaintTransform`, and hit-test through `addWithPaintTransform`.

## 3. `size` is constrained, the paint scale isn't

```dart
size = constraints.constrain(Size(width * _morphScale, _clusterHeight * _morphScale));
```

`constrain` can clamp, but `_paintChildren` scales unconditionally. A 3-icon cluster swells to ~168×56 — on a narrow device against a bounded row, children paint outside the box with no clip, over the title. Worth handling or at least asserting.

## 4. Per-frame layer + layout churn

- `pushTransform(true, …)` is called without `oldLayer:`, so a fresh `TransformLayer` is allocated every paint.
- A fresh `ImageFilterLayer` is constructed per blurring child per frame in `_paintChildren`.
- The `morphScale` setter calls `markNeedsLayout()` every frame, and `LiquidRoundedRectangle` lost its `const` so the glass shape/radius changes each frame too.

Net: every push/pop is now a full relayout plus a rebuilt layer subtree plus N blur layers per frame. For this package specifically that feels worth caching — store the layers in `layer`/parent data and reuse via `oldLayer:`, and consider `alwaysNeedsCompositing => true` rather than forcing compositing from inside `paint`.

## 5. The width doesn't overshoot, contrary to the description

`widthT` is `springT.clamp(0.0, 1.0)`. Tracing it, `clampedT` reaches 1.0 at p≈0.47 and stays pinned for the remaining **53%** of the transition — width, item positions, height lerp and stretch all freeze early, and everything after that is the ±3% uniform scale. That may well be the intended feel (the test asserting monotone width confirms it's deliberate), but the CHANGELOG entry is user-facing and says width and item positions "ride the spring… overshooting the target and settling back". Could that wording be corrected?

Related, minor: `GlassNavMorphCurve._terminal` is `simulation.x(0.5)` = **1.00932**, not 1 — the spring hasn't settled at the perceptual duration, so the curve is scaled down ~1% throughout. Harmless, but the constants were tuned against a slightly-off curve.

## Smaller things

- **The description advertises a fix that isn't in the diff.** It lists "a one-frame flash of the destination chrome at push start" as included, but the merge commit says *"Upstream's first-frame proxy-animation fix supersedes ours"*. Worth trimming so the changelog archaeology stays honest.
- **`GlassNavPinnedMetrics.crossFadeStart/End` changed units, not just values** — 0.3/0.7 → 0.25/0.6, but now a fraction of the morph window rather than of route progress. That class is exported from the barrel and documented as the alignment contract for non-`GlassAppBar` chrome, so a meaning change probably wants a CHANGELOG line.
- **`GlassNavPinnedState` gained a required ctor param.** Not in the barrel's `show`, but it is the return type of the public `GlassNavigationShellState.resolveState()` and the file is deep-importable, so it's technically breaking.
- **`dart format` is dirty** on both changed files (`glass_nav_pinned_host.dart`, `glass_nav_morph_test.dart`). No CI gate, but it'll churn the next diff.
- **Test nits:** `w.runtimeType.toString() == '_PinnedCluster'` silently matches nothing if the class is renamed, and `expect(scales.last, 1.0)` is exact float equality on an animated value.
- `outgoingSigmaAt` takes `morphT` while `incomingSigmaAt` takes raw route progress — inconsistent inputs for what reads as a symmetric pair.

—

**1** and **2** are the blockers for me; **3** and **4** I'd like a view on. The rest is bookkeeping. The core idea and the choreography are good — happy to take another look once these are addressed.

marco242424:
Hi @JakeThomson.

Thanks for the thorough pass — all addressed.

**1.** Fixed in `ffc4a90` (crossed with your review). The gel is now gated on the morph having real work to do, and the underlying signal was also misfiring: `crossFades` relies on const `Icon`s canonicalising to one instance, which empirically doesn't hold — so matched icons have been quietly cross-fading with themselves since 1.1.0. Icons are now compared by value; a trace test pins that an unchanged cluster neither pulses nor resizes.

**2.** Fixed in `5b1562f`: gel transform mirrored in `applyPaintTransform`, hit-testing via `addWithPaintTransform`, compositing declared through a dynamic `alwaysNeedsCompositing`.

**3.** Went with the assert: layout fails loudly in debug if the scaled size didn't get its room. In the bar it can't bite (unbounded `Row`), and the shell's `ClipRect` contains overflow.

**4.** Allocation half fixed in `5b1562f` (`oldLayer:` reuse + one `ImageFilterLayer` per glyph in a `LayerHandle`). The per-frame layout and non-const radius are the deliberate cost of real geometry: a paint transform over `useOwnLayer` glass clips at its ~12px texture headroom at a +22% swell — the shell/content disconnect this PR removes. Child constraints don't change per frame, so it's the cluster's own math re-running, not the subtree.

**5.** Wording fixed in changelog + description: the overshoot is intentionally not walked in width — it's the whole-shell squeeze. `_terminal` is deliberate normalisation to honour the Curve end-at-1.0 contract; the constants were tuned against this exact curve next to a native `NavigationStack`, so the ~1% is absorbed by calibration.

**Smaller things** — all done in `5b1562f`: flash claim removed (that fix is yours from #238), changelog line for the `crossFade` units change, both sigma helpers take `morphT`, `popping` optional, format clean, test nits fixed.

2,554 green under the CI selection. Ready for another look.

JakeThomson (approved):

sdegenaar:
Hey @marco242424. Amazing work and thanks for addressing all that and all the work on this. @JakeThomson thanks for reviewing as you were closest to this. Appreciate your guys time on this and look forward to further PR's and collaboration. Amazing work.

Cheers

PR #253, opened 2026-08-28
fix(AdaptiveGlass): blur: 0 no longer downgrades a surface to the frost fallback

### LINKED ISSUE #251: blur: 0 silently routes a premium surface to the frost fallback, losing the shader, rim and specular
setting `blur: 0` routes a premium surface to `_FrostedFallback`, so it loses the shader's rim and specular. the blur value is picking the renderer, and nothing in the api says it does.

```dart
// lib/widgets/shared/adaptive_glass.dart
if (quality == GlassQuality.minimal ||
    // Deliberately the *configured* blur, not the effective one: this
    // path is for surfaces asking for no frost, not for surfaces on
    // their way out.
    baseSettings.blur == 0 ||
    platformViewBackdrop) {
```

i can see the configured-vs-effective distinction is deliberate, so the routing may be intended. what surprised us is the cost: "no frost" also means no lit edge.

two identical premium surfaces, same rim and thickness, only the blur differs:

before

the `blur: 0` surface keeps a plain hairline from the decorations, but the rim and specular the control has are gone.

the layer already handles a zero blur on its own, pass 1 is guarded and skipped:

```dart
// liquid_glass_layer.dart
if (settings.effectiveBlur > 0) { ... } else {
  _blurLayerHandle.layer = null;
  _clipPathLayerHandle.layer = null;
}
```

so a premium surface with `blur: 0` renders refraction and lighting correctly if it reaches the layer:

after

found this over a platform view, where a blur achieves nothing because the backdrop can't be captured, so we wanted the rim without one.

pr up, exempting premium only so standard and minimal keep today's routing. happy to drop it for a doc line instead if the behaviour is intended.

flutter 3.47, impeller, ios 26, main at 26551ae.

Nixxx19:
closes #251

`blur: 0` routed a surface into `_FrostedFallback`, which draws no rim or specular, so asking for no frost also cost the surface its lit edge. the condition is now gone entirely, as you described: a zero blur no longer downgrades any tier.

safe because the renderer already handles it, pass 1 is guarded by `effectiveBlur > 0` and simply skipped, leaving refraction and lighting intact.

before, premium with `blur: 0` against a `blur: 6` control:

before

after:

after

`minimal` and `platformViewBackdrop` still route to the frost fallback, untouched.

one thing worth your eye, since your reasoning referenced the layer and the layer is the premium path: at `standard`, a `blur: 0` surface now reaches the lightweight shader rather than the frost fallback, so it renders but it does look different (middle capsule above). that is a visual change for anyone currently passing `blur: 0` at standard. it seemed right to follow what you asked rather than special-case premium, but say the word if you would rather scope it.

`flutter analyze` clean. verified on ios 26 simulator, impeller, flutter 3.47, against main at 26551ae.


Nixxx19:
@sdegenaar widened it as you said, the condition is gone entirely now so blur: 0 doesn't downgrade any tier. minimal and platformViewBackdrop still route to frost, untouched.

one thing worth a look before 1.2.0: your reasoning referenced the layer, and the layer is the premium path, so i checked standard too. a blur: 0 surface at standard now reaches the lightweight shader instead of the frost fallback. it renders fine, but it does look different, so it's a visual change for anyone passing blur: 0 at standard today. screenshot of all three tiers is in the pr body.

followed what you asked rather than special-casing premium, happy to scope it back if you'd prefer.

sdegenaar:
Hey @Nixxx19, good call flagging it — it's a real visual change at standard and worth documenting. The wider scope is the right answer though: blur: 0 should mean "no blur pass, keep the shader" at every tier. Anyone relying on blur: 0 to get _FrostedFallback at standard has an explicit alternative: GlassQuality.minimal. I'll note the visual behaviour change in the CHANGELOG. Merging — great work on both the issue and the fix. Appreciate it.

Cheers

PR #263, opened 2026-08-30
Scrollable segmented control: selection stability, anchored re-grid morph, and picker behaviors

jfhair:
The scrollable `GlassSegmentedControl` renders its indicator through a spring-follower that animates *every* value change, and its list-length path zeroes the indicator springs. Two visible artifacts follow: on every mount the pill slides in from the track edge to the selection, and on every list change it blinks out and slides back in. This PR fixes those and rounds the control out for picker-style use:

- **`VelocitySpringBuilder.teleportEpoch`** — a monotonic marker for discontinuous corrections (initial measure, list-change re-measure); the follower snaps instead of animating. Taps and drags animate exactly as before.
- **List-length changes keep the springs' geometry and the pill mounted**; the post-measure epoch-snap lands it with no travel.
- **`GlassSegment.id`** — cells are keyed by identity (falling back to label), so values that survive a list change keep their elements instead of remounting.
- **The selection scrolls into view** after initial and post-change measures (jump, not animate), and **`scrollController` is exposed**.
- **Anchored re-grid morph** (`regridDuration`, default 180ms): when the list is reconfigured around a surviving selection, entering segments grow in (width via `Align.widthFactor`, with scale 0.8→1 and fade riding the curve), leaving segments shrink out, and the survivors glide — with the selected cell held to its exact screen position by an analytic per-tick scroll offset computed from natural widths (never chasing last frame's boxes, so it stays correct even when a slow build paints only a few morph frames; the clock starts only after the pixel-identical t=0 frame absorbs the re-parenting cost). `Duration.zero` and Reduce Motion snap.
- **`SegmentSelectionAlignment.center`** — the picker policy: mount, tap, and re-grids all seat the selection mid-viewport (default `minimal`, unchanged).
- **`SegmentDragBehavior.scroll`** — every drag navigates the list and selection is tap-only (default `selectIndicator`, unchanged; fixed mode untouched). With centering, the selection parks exactly where a scrolling thumb naturally lands, so claiming that drag for the indicator was a usability trap on long lists.
- **Two-beat centered seating**: the pill travels to the tapped choice first while the scroll holds, then a 150ms-delayed 300ms glide centers it — selection feedback at the finger, then the world settles.

One default changes: list-length changes now morph (180ms) instead of snapping. If you'd rather keep the old rendering as the default, happy to flip `regridDuration` to `Duration.zero`.

Ten tests in `test/scrollable_selection_stability_test.dart`, including frame-stepped assertions that the rendered pill never moves through a morph and its finish. Full suite green apart from the five pre-existing golden drifts (identical on clean `main`).

Context: we built a time-of-day picker with 24–96 segments and re-griddable granularity (hour / 30-min / 15-min steps) on this control; happy to share screen recordings of before/after.

### DIFF
--- a/example/lib/demos/glass_segmented_regrid_demo.dart
+++ b/example/lib/demos/glass_segmented_regrid_demo.dart
@@ -1,0 +1,189 @@
+// Scrollable segmented control — re-grid morph + selection seating
+//
+// The scrollable control's hard case is a list whose LENGTH changes while a
+// selection is showing: a time picker that re-steps 60 → 30 → 15 minutes
+// re-grids every cell around the value you are looking at.
+//
+// Toggle "Morph" to compare the two candidate defaults for regridDuration:
+//   off (Duration.zero) — the list swaps instantly
+//   on  (180ms)         — entrants grow in, leavers shrink out, and the
+//                         selection stays anchored under your eye
+//
+// "Remount" rebuilds the control from scratch to show the seating beat: a
+// fresh mount lands with the selection already centred rather than sliding
+// in from the first segment.
+
+import 'package:flutter/cupertino.dart';
+
+import 'package:liquid_glass_widgets/liquid_glass_widgets.dart';
+
+Future<void> main() async {
+  WidgetsFlutterBinding.ensureInitialized();
+  await LiquidGlassWidgets.initialize();
+  runApp(LiquidGlassWidgets.wrap(child: const SegmentedRegridApp()));
+}
+
+class SegmentedRegridApp extends StatelessWidget {
+  const SegmentedRegridApp({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return const CupertinoApp(
+      title: 'Segmented — Re-grid Morph',
+      theme: CupertinoThemeData(brightness: Brightness.dark),
+      home: SegmentedRegridHome(),
+      debugShowCheckedModeBanner: false,
+    );
+  }
+}
+
+class SegmentedRegridHome extends StatefulWidget {
+  const SegmentedRegridHome({super.key});
+
+  @override
+  State<SegmentedRegridHome> createState() => _SegmentedRegridHomeState();
+}
+
+class _SegmentedRegridHomeState extends State<SegmentedRegridHome> {
+  /// Minutes-per-step, coarse → fine.
+  static const List<int> _steps = [60, 30, 15];
+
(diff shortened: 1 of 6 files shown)


sdegenaar:
Hey @jfhair, great work — teleportEpoch is the right fix for the mount-slide problem.

Targeting this for 1.2.2/1.30 (1.2.1 is shipping today). Two things before merge:

Could you share those before/after recordings? The re-grid morph and two-beat seating are choreography calls be good  to see in motion.

On regridDuration default — it's a grey area. Native iOS 26 UISegmentedControl doesn't scroll at all, so there's no platform reference to anchor the decision. Our instinct is Duration.zero by default with the morph as opt-in, but the video may change that.

Appreciate it.

Cheers

jfhair:
Thanks — and agreed on `regridDuration`. Pushed: the default is now `Duration.zero`, so the morph is strictly opt-in. Your reasoning holds — there's no platform behaviour to point at since native segmented controls don't scroll, and a package shouldn't start animating a host's control on upgrade. The two morph tests now opt in explicitly as well; without that the handoff one would have passed vacuously, having no morph to hand off.

Two recordings attached, from a new example demo (`example/lib/demos/glass_segmented_regrid_demo.dart`) — a time picker re-stepping 60 → 30 → 15 minutes, with a switch to compare the morph off and on against the same list, and a remount button for the seating beat.

https://github.com/user-attachments/assets/044075bd-8c54-40ac-a0e1-d41b8fa14f60

https://github.com/user-attachments/assets/54e5ed26-8ca9-4594-9ebc-f8df81a1791d

**Seating** — selecting a segment plays the two beats: the pill travels to the tapped cell while the scroll holds, then the strip glides that choice to centre with the pill riding its cell. Centering at the same time as the pill's travel made both land in one instant and nothing ever visibly arrived where the finger was; the 150 ms offset is what makes the order legible.

**Re-grid** — the same list re-stepping with the morph off, then on. Off, every cell swaps at once and the value you were reading is somewhere new. On, entrants grow in, leavers shrink out, and the selection holds its position.

Fair warning that the captures don't convey the feel as well as running the demo does — these are simulator recordings, and the morph in particular is brief enough that it reads better on device than in a video.

Happy to split the demo into its own PR if you'd rather keep this one to the widget changes.

sdegenaar:
Thanks @jfhair. Released in v1.2.2. Appreciate the contributions.

Cheers

PR #268, opened 2026-09-01
perf: optimize percentile with Quickselect and fix transform tracking crash

kdbhalala:
### Summary of Changes

As discussed in the review, this PR is now scoped strictly to the two non-controversial, high-value fixes:

1. **(N)$ Quickselect for Frame Timing Percentile Calculation:**
   * In `lib/utils/glass_quality_adapter.dart`, replaced (N \log N)$ full sorts on the frame timing callback path with in-place (N)$ average time Quickselect (`_quickSelect`).
   * Eliminates heap allocations on Phase 3 hysteresis checks and reduces algorithmic complexity for runtime P95 evaluations.
   * Added fast, lightweight algorithmic unit tests in `test/utils/glass_quality_adapter_test.dart` covering sorted, reverse-sorted, duplicate, and random arrays across all percentiles (runs in < 10 ms).

2. **Transform Tracking Crash Guard During Route Transitions:**
   * In `lib/src/renderer/internal/transform_tracking_repaint_boundary_mixin.dart`, added `renderObject.attached` check and `try-catch` around `renderObject.getTransformTo(null)`.
   * Prevents `StateError` crashes when transform callbacks fire while render objects are unmounting during page pops and route transitions.

---

### Reverted Changes as Requested:
* **Shaders:** Reverted changes to `liquid_glass_final_render.frag` and `interactive_indicator.frag` to preserve 12-tap `textureBilinear` for Flutter's nearest-neighbor backdrop filtering workaround (#139887).
* **Static Paint:** Reverted static final `Paint` instance caching in `adaptive_glass.dart` to avoid retaining native shader handles across the app lifecycle.
* **Microbenchmarks:** Removed 1M-iteration test files and standalone demo harnesses from `test/` to keep standard CI runs fast.

---

### Verification:
* **All 2,849 tests passing:** `flutter test` passes cleanly in ~1 minute.
* **0 Analyzer Issues:** `flutter analyze` reports no issues.

### DIFF
--- a/lib/src/renderer/internal/transform_tracking_repaint_boundary_mixin.dart
+++ b/lib/src/renderer/internal/transform_tracking_repaint_boundary_mixin.dart
@@ -87,7 +87,13 @@ class GeometryTransformTrackingLayer extends OffsetLayer {
 
   @override
   void addToScene(ui.SceneBuilder builder) {
-    final currentTransform = renderObject?.getTransformTo(null);
+    if (renderObject == null || !renderObject!.attached) return;
+    Matrix4? currentTransform;
+    try {
+      currentTransform = renderObject!.getTransformTo(null);
+    } catch (_) {
+      return;
+    }
     if (!MatrixUtils.matrixEquals(currentTransform, _lastTransform)) {
       // Don't trigger onTransformChanged on the very first frame (when _lastTransform is null).
       // The render object just painted itself, so it is already up to date. Triggering it
--- a/lib/utils/glass_quality_adapter.dart
+++ b/lib/utils/glass_quality_adapter.dart
@@ -751,7 +751,7 @@ class GlassQualityAdapter {
   // ── Percentile math ───────────────────────────────────────────────────────
 
   /// Computes the [percentile]-th percentile (0–100) from a list of integer
-  /// values. The list is not modified; a sorted copy is made internally.
+  /// values. The list is not modified.
   ///
   /// Uses the "nearest rank" definition — appropriate for frame timing data
   /// where we care about real observed samples, not interpolated values.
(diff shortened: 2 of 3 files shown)


sdegenaar:
Hey @kdbhalala , solid contribution — the Quickselect for percentile calculation is a real win, that path runs on every frame timing callback. The transform tracking crash fix is also good, we've seen that StateError in the wild during route transitions.

A few things need addressing before we can merge though:

Shader changes — please revert liquid_glass_final_render.frag and interactive_indicator.frag. textureBilinear is a deliberate workaround for Flutter's backdrop nearest-neighbor texture filtering (#139887). Dropping to raw texture() for the R and B channels will reintroduce visible edge aliasing on Retina screens during motion. The bandwidth saving isn't real in practice — those UV coordinates are within the same GPU cache line anyway.

Static Paint — revert this too. Mutating a static final Paint with ..shader = gradient keeps a reference to the native shader alive for the entire app lifetime. Per-frame Paint() allocations in Dart AOT are ~30ns and die in nursery GC — not worth the risk.

FPS numbers & benchmarks — the benchmark is summing FrameTiming.totalSpan per frame and dividing into frame count, which gives 1/latency not actual FPS. That's where the 490 figure comes from, not real throughput. Also, let's keep 1M-iteration microbenchmarks out of test/ so standard CI test runs stay fast.

If you want to get the good stuff landed fast, happy to take the Quickselect + crash fix as a smaller PR.

Thanks again and for using the package and contributing. Appreciated!

Cheers

kdbhalala:
Thanks for the thoughtful and detailed review! Makes total sense on all fronts:

1. **Shaders:** Reverted both `liquid_glass_final_render.frag` and `interactive_indicator.frag` to keep `textureBilinear` intact.
2. **Static Paint:** Reverted back to per-frame allocation so native shader references aren't held alive.
3. **Benchmarks:** Removed the 1M-iteration test file and benchmark demo from the PR entirely to keep standard CI tests fast.
4. **Focused Scope:** Kept strictly the **Quickselect (N)$ optimization** + **transform tracking route transition crash fix**, plus lightweight unit tests in `test/utils/glass_quality_adapter_test.dart` that run in a few milliseconds.

Branch is rebased and ready for review! 🚀

PR #294, opened 2026-09-07
fix(GlassNavigationShell): defer animation ticks that land in the build phase

### LINKED ISSUE #293: [Bug]: GlassNavigationShell calls setState during build on page-based navigations with a zero-duration route (1.4.0)
### What's happening?

With `GlassNavigationShell` installed above a page-based `Navigator` (go_router), every push or pop of a route that has no transition throws:

```
setState() or markNeedsBuild() called during build.
This ListenableBuilder widget cannot be marked as needing to build because the framework is already in the process of building widgets. ...
The widget on which setState() or markNeedsBuild() was called was: ListenableBuilder
The widget which was currently being built when the offending call was made was: HeroControllerScope
```

with the shell in the stack:

```
#3  _AnimatedState._handleChange
#4  ChangeNotifier.notifyListeners
#5  _TickNotifier.notify                 (glass_navigation_shell.dart:682)
#6  GlassNavigationShellState._onAnimationTick  (glass_navigation_shell.dart:306)
#7  AnimationLocalListenersMixin.notifyListeners
...
#10 AnimationController._animateToInternal
#11 AnimationController.reverse
#12 TransitionRoute.didPop
...
#16 NavigatorState._updatePages
#17 NavigatorState.didUpdateWidget
```

and, on a push, the same tick reached from `ProxyAnimation.parent=` inside `TransitionRoute._setSecondaryAnimation` → `didChangeNext` → `_updatePages`.

**Cause**

`GlassNavigationShellState` listens to each registered route's `animation` and `secondaryAnimation` and re-resolves the pinned chrome on every tick through a `ChangeNotifier` (`_TickNotifier`) that a `ListenableBuilder` in the shell's overlay rebuilds on. `_onAnimationTick` notifies synchronously.

A page-based `Navigator` applies its pages inside `didUpdateWidget`, which is the build phase. Two things there notify value listeners synchronously:

- a route with `transitionDuration: Duration.zero` completes its animation immediately in `_animateToInternal`, so `didPop` → `reverse()` ticks during the build;
- wiring a route's secondary animation (`ProxyAnimation.parent=`) notifies when the new parent's value differs from the proxy's, which it does when the neighbour is a completed zero-duration route.

The chrome's `ListenableBuilder` is a sibling of the `Navigator`, not a descendant of the rebuilding `Router`, so the mark is refused. The shell already defers *status* changes for this exact reason (`_scheduleNotify`), but not value ticks.

The 1.4.0 test commit removed "redundant secondaryAnimation tick listeners in GlassMenu and GlassPopover" for zero-duration route dismissal; the shell has the same problem.

### Steps to Reproduce

A widget test (included in the PR) fails on 1.4.0 and passes with the fix: a root-level `Navigator(pages:)` under `GlassNavigationShell`, two `GlassAppBar.pinned` screens, a `PageRouteBuilder` page with `transitionDuration: Duration.zero` pushed and popped by mutating the pages list from a `State` **below** the shell. `tester.takeException()` returns the error above after the pump that applies the pages.

### Versions

liquid_glass_widgets 1.4.0 (hosted), Flutter 3.47.1 stable, iOS 26 simulator (iPhone 17), Impeller. Found in a go_router app whose tab roots are `NoTransitionPage`s under the shell, so every tab switch hit it.

### Platform

iOS

### Rendering Backend

Impeller

### Anything else?

**Fix**

Route `_onAnimationTick` through the existing `_scheduleNotify()` so a tick that lands in `persistentCallbacks` / `midFrameMicrotasks` is queued to a post-frame callback (coalesced), and any other tick notifies inline as before. One line plus the test. A PR with the one-line change and the test follows.

JakeThomson:
## Description

`GlassNavigationShell` listens to every registered route's `animation` and `secondaryAnimation` and re-resolves the pinned chrome on each tick through a `ChangeNotifier` that the chrome's `ListenableBuilder` rebuilds on. `_onAnimationTick` notified synchronously.

A page-based `Navigator` applies its pages inside `didUpdateWidget`, which is the build phase, and two things there notify value listeners synchronously: a route with no transition completes its animation at once in `_animateToInternal`, so a pop's `reverse()` ticks mid-build, and wiring a route's secondary animation (`ProxyAnimation.parent=`) notifies when the neighbour's value differs from the proxy's, which it does when that neighbour is a completed zero-duration route. The chrome's `ListenableBuilder` is a sibling of the `Navigator`, not a descendant of the rebuilding `Router`, so the mark is refused: `setState() or markNeedsBuild() called during build` on every such push and pop, and the chrome misses that frame. In a go_router app whose tab roots are `NoTransitionPage`s, that is every tab switch.

The shell already defers status changes past build for this reason (`_scheduleNotify`), but not value ticks. This routes `_onAnimationTick` through the same deferral: a tick that lands in `persistentCallbacks` or `midFrameMicrotasks` is queued to one coalesced post-frame notification, and any other tick notifies inline as before.

The new test mounts a root-level `Navigator(pages:)` under the shell with the page state held **below** the shell (where go_router's `Router` sits), pushes and pops a `PageRouteBuilder` page with `transitionDuration: Duration.zero`, and asserts no exception after the pump that applies the pages. It fails on 1.4.0 with the error above and passes with the change. The 1.4.0 test commit removed the same kind of synchronous tick listener from `GlassMenu` and `GlassPopover`; the shell had the same problem.

Fixes #293

## Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)

## Platforms Tested

- [x] iOS (Impeller)
- [ ] Android (Impeller)
- [ ] macOS
- [ ] Windows
- [ ] Web
- [ ] Linux

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.4.1
+
+## Bug Fixes
+
+- **`GlassNavigationShell` no longer marks its chrome dirty mid-build (#293):** The shell listens to every registered route's animations and re-resolved the pinned chrome synchronously on each tick. A page-based `Navigator` applies its pages inside `didUpdateWidget` — the build phase — and a route with no transition completes its animation right there, so the tick landed as `setState() or markNeedsBuild() called during build` on the chrome's `ListenableBuilder` for every such push and pop, and the chrome missed that frame. Ticks now defer past build the way status changes already did.
+
 # 1.4.0
 
 ## Features
--- a/lib/widgets/surfaces/glass_navigation_shell.dart
+++ b/lib/widgets/surfaces/glass_navigation_shell.dart
@@ -303,7 +303,11 @@ class GlassNavigationShellState extends State<GlassNavigationShell>
     animation.removeStatusListener(_onAnimationStatus);
   }
 
-  void _onAnimationTick() => _tick.notify();
+  // A route wiring its secondary animation (`ProxyAnimation.parent=` inside
+  // `didChangeNext`) notifies value listeners synchronously, and for a
+  // page-based Navigator that runs inside `didUpdateWidget` — the build
+  // phase — so the tick has to defer exactly as a status change does.
+  void _onAnimationTick() => _scheduleNotify();
 
   void _onAnimationStatus(AnimationStatus status) {
     // Not while a finger is down: several routes' animations are listened to
--- a/test/widgets/surfaces/glass_navigation_shell_test.dart
+++ b/test/widgets/surfaces/glass_navigation_shell_test.dart
@@ -1065,6 +1065,55 @@ void main() {
     });
   });
 
+  group('page-based navigators', () {
+    testWidgets(
+        'a pages update that drives a route animation does not mark '
+        'the chrome dirty mid-build', (tester) async {
+      // A declarative Navigator applies its pages inside didUpdateWidget —
+      // the build phase. Wiring a secondary animation there
+      // (ProxyAnimation.parent=) and starting a pop's reverse both notify
+      // value listeners synchronously, so the shell's tick has to defer or
+      // the chrome's ListenableBuilder is marked dirty mid-build.
+      final key = GlobalKey<_PagesNavigatorState>();
+      await tester.pumpWidget(_PagesApp(navigatorKey: key));
+      await settle(tester);
+
+      key.currentState!.push(const _Screen(title: 'Detail', actions: []));
+      await tester.pump();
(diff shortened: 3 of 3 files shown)


sdegenaar:
Good catch on the didUpdateWidget tick race with zero-duration pages. Swapping _onAnimationTick over to _scheduleNotify() makes total sense—everything else in the shell was already routing through it anyway.

Tests look great. Merging this into 1.4.1 alongside another fix. 

Cheers

PR #297, opened 2026-09-08
fix(GlassNavigationShell): dissolve a custom item's own glass through its surface, not a paint layer

### LINKED ISSUE #296: [Bug]: A GlassBarItem.custom that draws its own glass renders as its backer across a push or pop and snaps in on the last frame
### What's happening?

A `GlassBarItem.custom` whose child is itself a glass surface — a `GlassButton.custom` capsule with `background: GlassBarItemBackground.none`, so the shell draws nothing behind it — renders wrong for the whole of a push or pop under `GlassNavigationShell`. On a push the incoming icons arrive with no capsule around them (with a `backerColor` set, an opaque frosted pill instead) and the real glass snaps in on the last frame of the transition. On a pop the capsule vanishes on the first frame and the icons blur away over bare content.

**Before** (item hoisted from a `GlassAppBar.pinned` on a photo page, pushed from a page with a shared text action):

https://github.com/user-attachments/assets/53313544-fe22-474f-8c62-11bec2203f87

**Cause**

`_RenderPinnedCluster._paintChildren` in `glass_nav_pinned_host.dart` fades and blurs each item's content by painting it under an `OpacityLayer` (`pushOpacity`) while it cross-fades and an `ImageFilterLayer` while the glyph sharpens, on the stated assumption that "item contents are not glass". A backdrop-sampling surface painted inside either layer has no backdrop to sample, so it renders as its backer (or nothing) until the layers are gone. The blur sigma only reaches zero at `glyphSharpenEnd` (morphT 0.9), which is why the snap lands at the end rather than mid-way.

The stock example never hits it because no demo hoists a glass-bearing item: data items are drawn inside the shell's own capsule, and the `none` items in the demos are plain content. Mirroring the setup in the example (two temporary screens, a text action in a shared capsule pushing to a photo page whose only action is a `GlassBarItem.custom` carrying a `GlassButton.custom` capsule) reproduces it exactly, and bypassing the two paint layers renders the capsule correctly throughout.

### Steps to Reproduce

1. Root route: `GlassAppBar.pinned(actions: [GlassBarItem.custom(id: 'x', child: Text('Select'))])`.
2. Push a route whose bar is `GlassAppBar.pinned(actions: [GlassBarItem.custom(id: 'x', background: GlassBarItemBackground.none, child: GlassButton.custom(shape: LiquidRoundedRectangle(borderRadius: 22), useOwnLayer: true, onTap: () {}, child: Row(children: [Icon(share), Icon(ellipsis)])))])` over a bright photo background.
3. Watch the trailing corner during the push and the pop.

A widget test that pumps the push to mid cross-fade and inspects the item's nearest `GlassMaterializeScope` distinguishes the two paths (included in the PR).

### Versions

liquid_glass_widgets 1.4.0 (hosted) and 1.4.1 (main), Flutter 3.47.1 stable, iOS 27 simulator (iPhone 17), Impeller.

### Platform

iOS

### Rendering Backend

Impeller

### Anything else?

The same layers are what give a plain `none` item (an avatar, a badge) its fade, so they cannot simply be dropped for `none`. The shell needs to know the child is glass and hand the dissolve to the surface itself through `GlassMaterializeScope` — visibility for the glass, opacity and blur for the content inside it — which every package surface already honours. Note `GlassMenu` wraps its trigger in its own `GlassMaterializeScope`, so that scope has to sit below the menu wrapper to be the nearest one.

JakeThomson:
## Description

`_RenderPinnedCluster._paintChildren` fades and blurs each item's content by painting it under an `OpacityLayer` while it cross-fades and an `ImageFilterLayer` while the glyph sharpens, on the assumption that item contents are not glass. A `GlassBarItem.custom` whose child is itself a glass surface — a `GlassButton.custom` capsule with `background: none`, the shape an app takes when it wants its own capsule in the bar — has no backdrop to sample inside either layer, so it renders as its backer (or nothing) for the whole transition and snaps to glass on the last frame, where the blur sigma finally reaches zero. On a pop the capsule vanishes on the first frame.

This adds `GlassBarItemBackground.own`: like `none`, nothing is drawn behind the item, but the cluster dissolves it through the surface's own visibility instead of a paint-time layer. `_OwnGlassDissolve` wraps such an item in a `GlassMaterializeScope` carrying the cluster's opacity as `glassProgress` and `contentOpacity` and its blur as `contentSigma`, composed with any enclosing scope so a menu morph can still fade the trigger; the cluster paints it plain. It sits at the cluster child, below the `GlassMenu` wrapper's own resting scope, so it is the nearest one. Plain `none` content keeps the layer fade — it is what gives an avatar or badge its cross-fade — and the in-route fallback treats `own` as `none`.

The new tests push a route with an `own` item and assert that mid cross-fade its nearest `GlassMaterializeScope` is partway through with a non-zero sigma, and settled at one after the transition; and that a `none` item at the same moment sees only the resting scope. The first fails on main and passes with the change.

A second commit fixes what the same recording showed once the item was `own`: the outgoing cluster stayed solid until the shell dropped it, the incoming one appeared solid, and for a few frames both were drawn. `GlassMenu` wraps its trigger in a resting `GlassMaterializeScope` so the trigger can fade under an open menu, and every pinned cluster sits inside that wrapper — so the materialize the shell runs around a cluster only one route has never reached the glass. The menu's scope now composes with an enclosing one. A test pushes a route whose leading cluster exists only on that route and asserts its nearest scope is partway through mid-materialize; it fails on main.

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/a56f93ac-a4c3-4431-92f2-9e7348fec1d9" /> | <video src="https://github.com/user-attachments/assets/d75f96fb-f35e-4188-b3a2-38e4bf247efb" /> | 

Fixes #296

## Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)

## Platforms Tested

- [x] iOS (Impeller)
- [ ] Android (Impeller)
- [ ] macOS
- [ ] Windows
- [ ] Web
- [ ] Linux

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - **Custom bar items that draw their own glass no longer flash across a push or pop:** The pinned chrome fades and blurs item content by painting it under opacity and image-filter layers, and a glass surface painted under either has no backdrop to sample — a `GlassBarItem.custom` carrying its own `GlassButton.custom` capsule rendered as its opaque backer for the whole transition and snapped to glass on the last frame. `GlassBarItemBackground.own` marks such an item, and the cluster dissolves it through the surface's own visibility instead.
 
+- **Pinned clusters now dissolve across a push or pop instead of popping in and out:** `GlassMenu` wraps its trigger in a resting `GlassMaterializeScope` so the trigger can fade under an open menu, and every pinned cluster sits inside that wrapper — so the materialize the shell runs around a cluster only one route has never reached the glass. The outgoing shell stayed solid until it was dropped, the incoming one appeared solid, and for a few frames both were drawn. The menu's scope now composes with an enclosing one.
+
 # 1.4.1
 
 ## Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 1.4.2
+
+## Bug Fixes
+
+- **Custom bar items that draw their own glass no longer flash across a push or pop:** The pinned chrome fades and blurs item content by painting it under opacity and image-filter layers, and a glass surface painted under either has no backdrop to sample — a `GlassBarItem.custom` carrying its own `GlassButton.custom` capsule rendered as its opaque backer for the whole transition and snapped to glass on the last frame. `GlassBarItemBackground.own` marks such an item, and the cluster dissolves it through the surface's own visibility instead.
+
 # 1.4.1
 
 ## Bug Fixes
--- a/docs/GLASS_NAVIGATION_TRANSITION.md
+++ b/docs/GLASS_NAVIGATION_TRANSITION.md
@@ -136,7 +136,10 @@ bar.
   `separate` (`sharesBackground: NO` — its own shell, which at a lone icon's
   size is the circular button the back button already is), and `none`
   (`hidesSharedBackground` — no glass, for content that carries its own shape,
-  such as a profile photo).
+  such as a profile photo). A fourth, `own`, is `none` for content that is
+  itself a glass surface: the shell dissolves it through the surface's own
+  visibility instead of fading it under a layer, which a glass surface cannot
+  survive.
 - **`id` mirrors `UIBarButtonItem.identifier`.** Items sharing an `id` across
   two routes are treated as the same item and hold their position while
   everything around them morphs. Without an `id`, items are matched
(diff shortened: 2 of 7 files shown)


sdegenaar:
Hey @JakeThomson. Thanks again for all the PR's. Really appreciate the help!! This once was a nice polish fix :-)

Thanks again!!!!!!!

Cheers

PR #321, opened 2026-09-15
fix(engine): clamp the glow touch to the layer it lights

### LINKED ISSUE #320: [Bug]: the interaction glow slides off the layer it lights when the pointer leaves it
### Version

1.6.0

### What happens

The interaction glow leaves the surface it belongs to. Press a `GlassTabBar.bottom` and hold, then slide your finger off the bar — vertically is enough — and the touch light fades out and disappears while the finger is still down. Slide back over the bar and it returns.

The selection indicator does not do this. It keeps tracking the same finger along the bar, because the drag physics resolve the pointer against the bar's own box. So during one continuous gesture the light and the indicator disagree about where the finger is, and the bar looks like it lost the touch it is still following.

### Why

`GlassGlow._handlePointer` converts the pointer into the layer's coordinate space and passes it straight to `GlassGlowLayer.updateTouch`:

```dart
pos = layerBox.globalToLocal(myBox.localToGlobal(event.localPosition));
```

Nothing constrains `pos` to the layer. A captured pointer keeps reporting `onPointerMove` after it leaves the widget, so the glow centre walks outside, and `GlassGlowLayer` clips its paint to `clipper`. The overlap shrinks, the light dims, and past the glow radius there is nothing left inside the clip to draw.

It is most visible on a wide, short surface. `GlassGlow.glowRadius` is measured against the **shortest** side of the layer, so on a 64 x 391 bottom bar the radius comes from the 64. A vertical excursion of a little more than the radius empties the clip completely, and vertical is the easy direction to leave a bar that is only 64 tall.

The same position feeds `_touchSpecularNotifier`, so the shader's touch specular walks off with it, and that is clipped to the glass shape too.

### Expected

The highlight stays on the element while the press is alive. On the platform this mirrors, the light belongs to the control being pressed and parks against its edge rather than sliding off it.

### Reproducer

```dart
late BuildContext inner;
await tester.pumpWidget(
  MaterialApp(
    home: Center(
      child: SizedBox(
        width: 400,
        height: 64,
        child: GlassGlowLayer(
          child: GlassGlow(
            child: Builder(builder: (context) {
              inner = context;
              return const SizedBox.expand();
            }),
          ),
        ),
      ),
    ),
  ),
);

final notifier = GlassGlowLayer.touchSpecularNotifierOf(inner)!;
final gesture = await tester.startGesture(tester.getCenter(find.byType(GlassGlow)));
await tester.pump();
await gesture.moveBy(const Offset(0, -300));
await tester.pump();

// Actual:   -268.0
// Expected: within 0..64
expect(notifier.value.position.dy, inInclusiveRange(0, 64));
```

Seen on an Android emulator with a real app as well, not only in tests.

### Note on the radius, separately

The default `interactionGlowRadius` for `GlassTabBar.bottom` is `1.5`, and because the radius is normalised to the shortest side that is 96 on a 64-tall bar: a 192 wide disc on a 391 wide bar, so pressing one tab lights roughly half of it. `GlassTabBar.searchable` defaults to `1.0` for the same geometry. That is a separate question from this bug and I have not touched it in the fix — a host can already pass its own value — but the normalisation may be worth revisiting for surfaces whose long side is several times the short one.

azizibahram:
Fixes #320.

## The bug

`GlassGlow._handlePointer` converts the pointer into the layer's coordinate space and hands it to `GlassGlowLayer.updateTouch` unconstrained:

```dart
pos = layerBox.globalToLocal(myBox.localToGlobal(event.localPosition));
```

A captured pointer keeps reporting `onPointerMove` after it leaves the widget, and `GlassGlowLayer` clips its paint to `clipper`. So the glow centre walks outside the clip, the overlap shrinks, and past the glow radius there is nothing left inside to draw. The light dies while the finger is still down, and comes back if you slide in again.

It reads worst on a wide, short surface. `glowRadius` is measured against the **shortest** side of the layer, so on a 64 x 391 `GlassTabBar.bottom` the radius comes from the 64, and a vertical excursion of a little over the radius empties the clip. Vertical is also the easy direction to leave a bar that is only 64 tall. Meanwhile the selection indicator resolves the pointer against the bar's own box and keeps tracking the same finger, so one continuous gesture produces two different answers about where the touch is.

The same position feeds `_touchSpecularNotifier`, so the shader's touch specular walked off with it, and that is clipped to the glass shape too.

## The fix

Clamp once, where the layer box is already resolved:

```dart
pos = Offset(
  pos.dx.clamp(0.0, layerBox.size.width),
  pos.dy.clamp(0.0, layerBox.size.height),
);
```

The highlight parks against the nearest edge and keeps tracking along it, which is what the platform does — the light belongs to the element being pressed. Both the glow circle and the shader specular are fixed by the one change, since they share the position.

## Scope

- A press that never leaves its surface is unaffected: its centre is already inside, so `clamp` is the identity.
- The fallback branch (no attached render boxes, i.e. tests and mid-layout) is untouched, because there is no layer box to clamp against there.
- It applies to every `GlassGlow`, not only the tab bars. That is deliberate: the layer clips the glow, so a centre outside the clip is never useful — it can only fade the light out. If you would rather it were opt-in for `GlassButton` and friends, say so and I will put it behind a flag.

## Test

`test/renderer/glass_glow_test.dart` gains `GlassGlowLayer bounds > clamps the touch to the layer when the pointer leaves it`. It reads the position through the existing public `GlassGlowLayer.touchSpecularNotifierOf`, so it needs no new API.

Without the fix it reports `-268.0` for a pointer dragged 300 above a 64-tall layer. With it, the position stays inside on both axes, and the unclamped axis keeps tracking (`dx` still follows the finger while `dy` is pinned).

## Checks

- `flutter analyze lib test` — clean.
- `flutter test test/widgets test/utils test/renderer/glass_glow_test.dart` — 2449 passing.
- The wider `test/golden/**` and `test/renderer/**` image comparisons fail identically on a clean `main` on this host (font and shader drift), so I left them out of the gate above rather than report them as results of this change.

Also verified on an Android emulator in a real app: before, holding a tab and sliding off the bar killed the light mid-gesture; after, it parks on the bar's edge and keeps travelling with the finger.

## Separate, not in this PR

`GlassTabBar.bottom` defaults `interactionGlowRadius` to `1.5`, which on a 64-tall bar is a 96 radius, a 192 wide disc on a 391 wide bar — pressing one tab lights about half of it. `GlassTabBar.searchable` defaults to `1.0` for the same geometry. Hosts can already pass their own value, so this is not a bug, but the shortest-side normalisation may be worth revisiting for surfaces whose long side is several times the short one. Happy to open that separately if it is of interest.


azizibahram:
Withdrawing this for now — clamping the centre keeps the light alive but it is not the whole fix, and I would rather not land a half-answer. Leaving #320 open with the diagnosis; I will come back with a version that makes the highlight follow the element rather than the raw pointer.

azizibahram:
Reopened as #322 \u2014 the branch had been force-pushed, so this one could not be reopened in place.

PR #329, opened 2026-09-16
feat(tab-bar): let a null interactionGlowRadius reach the native glow

### LINKED ISSUE #323: [Bug]: GlassTabBar cannot reach the native specular blur that GlassButton documents (sigma 4 vs 16)
### Version

1.6.0

### Summary

`GlassButton` and `GlassTabBar` resolve two different touch specular models, and only the button's is the one the package documents as matching the platform. The bar cannot reach it, because the gate is a `null` radius and the bar's radius parameter is not nullable.

### The two models

`GlassButton` treats a null `glowRadius` as native mode:

```dart
// glass_button.dart
final effectiveGlowRadius = widget.glowRadius ?? 1.6;

// Resolve effective blur: null falls through to 16.0 for a creamy organic falloff.
final nativeGlowBlurRadius =
    widget.glowRadius == null ? 16.0 : effectiveGlowBlurRadius;
```

with a doc comment that spells out the intent:

> Native iOS 26 mode. Resolves to a wide, calibrated 1.6 radius with a soft sigma-16 Gaussian blur and subtle specular alpha, clipped strictly to `shape`. The light washes smoothly across the button surface without producing a concentrated, pointy hotspot, matching Apple's touch-reactive glass reflection.

`GlassTabBar` takes the theme's sigma unconditionally:

```dart
// tab_bar_bottom_layout.dart
// Glow appearance fields come from the theme; they cannot be set per-widget
// because they are part of the theme palette.
final effectiveGlowBlurRadius = resolvedGlowColors.glowBlurRadius;
```

and the theme's fallback is sigma 4:

```dart
// glass_theme_data.dart
// Appearance defaults — sigma-4 blur softens the glow edge for a natural
// liquid-glass feel.
glowBlurRadius: 4.0,
```

### Why the bar can never get the native one

`interactionGlowRadius` is a non-nullable `double` with a default on every factory, so there is no null for a native branch to key on:

| Factory | `interactionGlowRadius` |
| --- | --- |
| `GlassTabBar.bottom` | `1.5` |
| `GlassTabBar.searchable` | `1.0` |
| the other two | `1.5` |

(The `1.5` against `1.0` looks unintentional too, for what is otherwise the same geometry.)

### What it looks like

Sigma 4 across a wide, short surface is a disc with an edge on it rather than a sheen. It shows up most on `GlassTabBar.bottom`, because `GlassGlow.glowRadius` is measured against the layer's **shortest** side, so on a 64 x 391 bar the radius comes from the 64 and the light lands as a bright circle sitting on the bar.

Changing `interactionGlowRadius` does not help, which is the part that cost me the most time: at 1.5 it is a wide disc, at 0.75 a smaller one, at 0.4 a dot. The radius sets how big the disc is; only the sigma decides whether it is a disc at all.

Raising the theme's `glowColors.glowBlurRadius` to 16 fixes the bar immediately and looks like the button's documented native sheen. But the theme is global, so a host that wants one bar to match the platform has to change the falloff of every glow in the app to get it.

### Suggestion

Give the bar the same native path the button already has — for example make `interactionGlowRadius` nullable, with `null` resolving to 1.6 + sigma 16 + the native alpha, and keep the current values as the explicit non-null behaviour so nothing existing shifts. Exposing `interactionGlowBlurRadius` on the public factories would also work, though it cuts against the "part of the theme palette" comment above.

Happy to send a patch for whichever shape you prefer; I did not want to guess at it in a pull request, since either choice changes how existing bars look.

### Note

Unrelated to #320, which is about the glow's *position* leaving the layer. This one is about its *falloff*. A host hitting both sees a hard-edged disc that also vanishes when the finger slides off the bar.

azizibahram:
Fixes #323.

This is the shape you picked on the issue: `interactionGlowRadius` becomes nullable across every `GlassTabBar` factory, and `null` means native mode, the same calibration `GlassButton` resolves from a null `glowRadius`.

## A correction first

I had the 1.0 vs 1.5 discrepancy on the wrong factory. It is not `.searchable`. That one defaults to `1.5` like `.bottom` and `.minimizable`. The `1.0` is on **`GlassTabBar.inline`**, and its own comment calls it a "compact default suited for inline placement", so it was a choice, not a typo. I said `.searchable` in the "Separate, not in this PR" note on #322 and repeated it on the issue. Sorry for the noise. You were right not to assume.

## The change

- `interactionGlowRadius` is `double?` on `bottom`, `inline`, `searchable` and `minimizable`, and on the private constructor they delegate to. The default is `null`.
- The resolution lives in one pure function, `resolveTabBarInteractionGlow` in `tab_bar_layout_utils.dart`. `TabBarBottomLayout` and `TabBarSearchableLayout` both call it, so the two layouts can't drift apart again. Before this, each resolved the glow on its own.
- `null` resolves to radius `1.6` and blur `16`, the figures `GlassButton` uses.
- An explicit radius keeps the old behaviour exactly: the theme's `glowBlurRadius` and the theme's primary glow colour.
- An explicit `interactionGlowColor` still wins in both modes.

## One call you should make: the alpha

You named the radius and the sigma. I also carried over the **colour**, because `GlassButton`'s native branch does the same. In native mode it replaces the theme's adaptive primary with a lower white:

| | light | dark |
|---|---|---|
| theme adaptive primary (`glowColorsFor`) | `0x3DFFFFFF` (~24%) | `0x2AFFFFFF` (~16%) |
| `GlassButton` native sheen | `0x1AFFFFFF` (~10%) | `0x12FFFFFF` (~7%) |

The theme value is roughly double. At a `1.6` radius under a sigma-16 blur it reads as a haze over the pressed tab, not the soft wash the button gets. So I treated radius, blur and alpha as one calibration.

That changes one existing contract. `collapsed logo GlassButton receives GlassThemeData.primary glow color` built its `.searchable` bar with no radius, which now means native mode, so the theme's green never landed. I didn't weaken the assertion. The test now passes an explicit radius, which is still the path where the theme palette wins, and a new sibling test states the native behaviour.

If you'd rather native mode set the geometry only and left the colour to the theme, it's a one-line change in the resolver, and that test goes back to how it was. Tell me which you want.

## Visual check, as you asked

I ran it on an Android emulator (Impeller, OpenGL ES): `.bottom`, `.searchable` and `.inline`, each shown twice, with nothing passed (native) and with the old default passed explicitly. The bars used `GlassInteractionBehavior.glowOnly`, so there is no press scale and the indicator doesn't move. The only thing that changes under the finger is the light. I compared each held frame against the resting frame, pixel by pixel.

- **Old default** (`1.5` or `1.0`, theme sigma 4): a bright core with a visible falloff ring around the finger. This is the concentrated hotspot the `GlassButton` docs describe avoiding.
- **Native** (`1.6`, sigma 16, native sheen): dimmer and wider, a soft gradient with no core and no visible edge.

`.searchable` at 1.6 looks the same as `.bottom` at 1.6, so I see no reason to keep a separate default for it.

`.inline` at 1.6 reads the same way too. The radius is a fraction of the layer's shortest side, not a fixed size, so a short bar gets the same proportions. `GlassButton` uses 1.6 on controls smaller than an inline bar.

I have the before/after difference images and can attach them if useful.

## Behaviour change

`null` is the new default, so a bar that never passed a radius will look different after upgrading: `1.5`/sigma-4 (or `1.0` on `.inline`) becomes `1.6`/sigma-16 with the lower alpha. I put the CHANGELOG entry under **Features** with that spelled out, rather than under Bug Fixes, so nobody has to work out from a fix line that their bars changed.

## Tests

New file `test/widgets/surfaces/tab_bar_native_glow_test.dart`:

- five pure-Dart tests on the resolver: native mode whole, the light-mode sheen, an explicit radius keeping the palette, an explicit colour winning over native, and a null theme colour passing through.
- five widget tests: `.bottom`, `.searchable`, `.minimizable` and `.inline` each reach `1.6`/`16` with nothing passed, and `.bottom` with `0.75` still reaches the indicator at `0.75`/`4`.

`glass_tab_bar_searchable_test.dart` has the change to the theme propagation test described above, plus the new native-mode test.

## Checks

- `flutter analyze lib test`: clean.
- `flutter test test/widgets test/utils`: 2448 passing.
- `dart format` on the touched files.
- `dart doc --dry-run` crashes with a `RangeError` inside dartdoc 9.0.4's `_stripDocImports`. It crashes the same way on a clean `main` on this machine, so it isn't caused by this change.
- The `test/golden/**` and `test/renderer/**` image comparisons fail the same way on a clean `main` on this machine, as on #322, so I left them out of the gate above.


sdegenaar:
Merging as is. Option A.(full calibration): coupling radius (1.6), blur sigma (16), and the lower-alpha sheen together makes total sense visually and avoids the milky haze issue with the theme's primary color.

Thanks again for you quality work and contribution. Really helps.

Cheers

PR #337, opened 2026-09-19
fix(shaders): keep rim refraction inside the glass

Ahmadre:
Hi Sebastian,

we ran into small rainbow-coloured specks along the bottom edge of premium glass pills in our app, mostly under the left half of each pill. They show up whenever there's text close above the pill, like a page title right over a row of filter pills.

## Before / after

Pill row right under a large title (macOS, Impeller/Metal, light mode):

| Before | After |
| --- | --- |
| before | after |
| before, zoomed | after, zoomed |

Same thing in our compact header:

| Before | After |
| --- | --- |
| before | after |
| before, zoomed | after, zoomed |

## What's going on

At the rim the normal is almost horizontal, and the displacement from `refract()` in `liquid_glass_render.frag` grows with `uThickness` (`baseHeight = uThickness * 8.0`), not with the size of the shape. With our settings (thickness 30, refractive index 1.59, 2x display) the last rows inside the rim of a 36 pt pill end up 90 to 135 physical pixels away from where they are, which is more than the whole pill (72 px). So the bottom rim samples the backdrop above the pill's top edge, picks up the title there, and the chromatic dispersion splits it into red, green and blue. The offsets I measured in the screenshots (about 102, 110 and 127 px per channel) match that pretty closely.

## The fix

The displacement is capped at half of the geometry matte's shorter side, right after the pixel-ratio scale and before anything samples with it. That keeps every sample, dispersion offsets included, inside the surface's own footprint. Bars, sheets and other larger surfaces never get close to that limit, so they render the same as before.

I added `test/shaders/rim_refraction_reach_test.dart` as a source-level check, similar to `gles_flip_guard_test.dart`, since `flutter test` runs through Skia and never touches the premium shader.

Thanks a lot for this package 🫶. 

It's a big part of how our app looks and feels, and it's been great to build on.

Rebar


sdegenaar:
Hey @Ahmadre. Thanks for submitting the PR and the kind words. Great seeing it in action out there and your screens look good!!

Thanks again. Quality work

Cheers

PR #347, opened 2026-09-23
feat(LiquidGlassSettings): add the iOS 27 glass material for premium quality

### LINKED ISSUE #346: [Feature]: An iOS 27 material for premium glass — the frost, outline, rim light and folded lens band of glassEffect(.regular)
### What kind of feature?

New property or option for an existing widget

Premium glass reads as refracted, tinted glass, but beside a SwiftUI `glassEffect(.regular)` control on the same screen it is visibly a different material. Four things account for most of the difference, measured pixel for pixel with the two side by side:

- **The frost.** Native glass is a heavy cloud (a σ of about 14 pt on a 56 pt control) through which a lightly blurred copy of the content still shows at about a quarter strength, so text under a button reads as pale ghosts rather than going out of focus. That copy is held asymmetrically: through the light material black detail never goes more than a fixed step darker than the cloud, so a black stripe is a flat pale band and a white one a narrow hump, and dark content reads bolder; the dark material is the reverse. The cloud itself is not a plain mean either — over black-on-white text it stays near white.
- **The outline.** A half-point hairline on the very edge: the unblurred backdrop just outside the shape, darkened by a fixed step (about 31% across the light axis, a fifth of that at its ends). With the rim light below, it is most of what reads as the native "bubble" edge.
- **The rim light.** Two sharp lobes at the ends of the light axis, one pixel inside the outline, with a soft tail into the body.
- **The lens band.** Native refraction is identity out to 0.6 of the radius and then folds, showing an evenly compressed, mirrored copy of the interior (1.9:1 from 0.27–0.72 of the radius). Exact Snell refraction through the bevel folds much harder toward the edge.

None of these can be reached with the current settings, so an app that wants its chrome to sit beside system controls cannot get there by tuning.

| Native Light | Default Light | Native Dark | Default Dark |
|---|---|---|---|
| Image | Image | Image | Image |

New `LiquidGlassSettings` terms for each, all off by default so existing settings render exactly as they do now, and two presets tuned to the native light and dark materials:

```dart
GlassButton(
  settings: isDark ? LiquidGlassSettings.ios27Dark : LiquidGlassSettings.ios27Light,
  quality: GlassQuality.premium,
  ...
)
```

- `frost`, `frostOpacity`, `frostClamp`, `frostWeight` — the cloud, how much of the content shows through it, the side it is held on, and how much more light (or dark) pixels weigh in the cloud; `blurWeight` does the same for the copy.
- `rimShade`, `rimShadeEnds`, `rimLight` — the outline and its highlight.
- `lensModel: GlassLensModel.paraxial` — small-angle refraction, which gives the even fold.

The frost has to be cheap enough for chrome: several buttons over a scrolling page at 120 Hz. A frost built as extra `BackdropFilter` stages composed around the blur costs roughly a full-screen pass each on Impeller, so it needs to be one plain blur that the glass shader combines with the sharp backdrop itself.

I realise this is a large addition to the premium path, and that its look is a matter of taste as much as measurement, so the shape of it is completely up for discussion. In particular:

- **Opt-in presets or a new default.** As sketched it is opt-in: nothing changes unless an app asks for the presets. If you would rather the native material became the premium default, the current look could stay available as its own constants instead.
- **Where the switch lives.** Static presets on `LiquidGlassSettings` suit a per-widget choice. An app-wide switch could be `GlassThemeVariant` presets beside `light`, `dark` and `minimal`, which would mean `GlassThemeSettings` carrying the new terms.
- **The API surface.** Nine new settings is a lot. Some could be internal constants behind the presets if you would rather expose only the presets.

Tuning the existing settings: blur, tint, saturation, thickness, `ambientRim` and `fresnelStrength` get the body tint and saturation close, but no combination produces the ghosted text, the outline or the even fold. A frost built from the existing `blur` alone is either out of focus or not frosted.

### Anything else?

_No response_

JakeThomson:
## Description

Adds the four things that separate premium glass from a SwiftUI `glassEffect(.regular)` control on the same screen (#346): the frost, the hairline outline, the rim light and the folded lens band. Each is a new `LiquidGlassSettings` term, off by default, and `LiquidGlassSettings.ios27Light` / `ios27Dark` put them together as measured against the native light and dark materials. With the new terms at their defaults every existing surface renders pixel-identically to 1.7.2 — checked by diffing a screen of default premium buttons, capsules and text across the two builds.

This is a large change to the premium path, and whether it should stay opt-in, become the premium default with the current look kept as constants, or move to a theme-level switch is entirely open — the options are laid out in #346, and I'm happy to reshape any of it.

- **Frost** (`frost`, `frostOpacity`, `frostClamp`, `frostWeight`, `blurWeight`). One blur `BackdropFilterLayer`, clipped to the shape and to every odd physical pixel row of it, so the sharp backdrop survives on the rows between. The render shader reads the cloud from the odd rows and the sharp backdrop from the even ones, and builds the rest in its one pass: a luminance-weighted Gaussian of the sharp rows for the copy of the content that shows through, held within `frostClamp` of the cloud on one side, mixed at `frostOpacity`. The rows are built in pass-physical pixels and mapped back through the layer's transform, so they stay on whole rows at any position, scale and enclosing pass; a rotated or skewed surface and the capture path get no frost, and the neck between blended shapes, outside every shape's own path, stays unfrosted. A `blur` wider than the shader's 45-tap ghost takes (about 0.8 pt at 3×) runs as the existing blur pass first. `frostWeight` other than 1 adds a `ColorFilter.matrix` pass (`BlendMode.dstIn`) that writes a luminance weight into alpha, so the blur's unpremultiplied result is a weighted mean and the sharp rows unpremultiply back unchanged; the weight only covers the part of the blur inside the shape, so it tells more on large surfaces than on the 56 pt control it was measured on. This shape is what made it affordable: composing any extra stage around the backdrop blur (tone curves, a dilate, a mask) measured on an iPhone 17 Pro Max at roughly a full-screen pass each, and a first version built that way ran at 10 fps with eight buttons. A lone morphology filter and anything whose odd and even rows differ in what the blur reads were also ruled out: Impeller's large-sigma blur downsamples with a phase that depends on screen position.
- **Outline and rim light** (`rimShade`, `rimShadeEnds`, `rimLight`). The line is the unblurred backdrop just outside the shape, averaged along the edge and darkened by a fixed step — subtracted, not blended, so it stays the same step below any colour. With `rimShade` or `frost` above zero the geometry pass also draws the native edge (`uNativeEdge`): anti-aliasing narrowed to half a logical pixel so the line and the frost reach an opaque pixel, and the silhouette grown by a physical pixel where the native outline sits. Without either the geometry is exactly as before, including the wider edge that hides stair-stepping under the press scale.
- **Lens** (`GlassLensModel.paraxial`). The thin-prism law with the slope taken as sin²: identity out to 0.6 of the radius, then an even 1.9:1 fold, as line probes measure on the native band. `spherical`, the existing refraction, stays the default.

Measured on an iPhone 17 Pro Max (profile, eight standalone premium buttons over a page scrolling at 700 pt/s, raster time per frame, two runs):

| Recipe | Raster ms | Frame rate |
|---|---|---|
| Existing premium recipe | 4.29 / 4.19 | 120 fps |
| `ios27Light` | 3.56 / 3.55 | 120 fps |
| `ios27Dark` | 2.09 / 1.71 | 120 fps |
| `ios27Light` with `frostWeight: 1` | 3.08 / 3.10 | 120 fps |

Against the native control (iOS 27 iPhone 17 simulator, same geometry and backdrop): light stripes 191/221 against 192/218, light text plateau and ghost floor 244/230 against 247/236, dark stripes 98/132 against 96/129, dark text plateau and ghost peak 46/62 against 44/63. The dark material over a white page is not reproduced — native greys to about 184, this stays white — and a pressed light button goes almost white, the existing touch glow adding onto the bright frost. The harness that took these (a SwiftUI host beside `GlassButton`) is not included, as it needs a UIKitView factory in the untracked `example/ios`.

Tests: the new settings through defaults, copyWith, lerp, equality, `copyWithPinch` and a `GlassThemeSettings` override; the frost rows driven through the render object directly (the glass pass is gated on `ImageFilter.isShaderFilterSupported`), checking every odd physical row and no even one is in the clip at a fractional offset, under a 1.3× scale and inside a `BackdropFilter` whose top sits on an odd row, and that a rotated surface or no frost has none; and the two new uniforms pinned to the last slot of each shader.

Closes #346.

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [x] 🎨 Visual / shader change (impacts rendering output)

## Platforms Tested

- [x] iOS (Impeller)
- [ ] Android (Impeller)
- [ ] macOS
- [ ] Windows
- [ ] Web
- [ ] Linux

## Checklist

- [x] If this changes shader code, I have verified it compiles on both Metal (macOS) and SkSL/SPIR-V (Windows) — with `impellerc --sksl` and the Metal, GLES and Vulkan runtime stages
- [x] No dynamic indexing or unsupported GLSL builtins were introduced in shader code

## Screenshots / Recordings (if applicable)

Native / 1.7.2 default premium / this PR's presets:

| Light | Dark |
|--------|--------|
| 01_light | 02_dark |

| Text through the frost, light (3×) | Text through the frost, dark (3×) |
|--------|--------|
| 03_light_text_zoom | 04_dark_text_zoom |

| Stripes | Sizes |
|--------|--------|
| 05_stripes | 06_sizes |

| Greys | Colours | Dark and Greys |
|--------|--------|--------|
| 08_greys | 09_colours | 10_dark_greys |


sdegenaar:
Hey @JakeThomson , this is outstanding work!! Well done. The interleaved-row frost architecture is the only viable approach here given Impeller's cost model, and the fact that ios27Light actually ends up cheaper than stock premium because the main blur pass gets skipped is a fantastic result. The native measurement fidelity on the line probes is impressive, the cross-platform CI results are clean, and the backward-compatibility story is airtight. Since you've marked this as a draft and explicitly invited discussion on the open questions, here are our thoughts:

**Opt-in vs. Premium Default vs. Theme Switch**

Keep opt-in as you have it. Making ios27Light the new default for GlassQuality.premium would be a visible breaking change for every app tuned for 1.x behaviour — apps would shift without asking for it. The ios27Light / ios27Dark preset approach is exactly right, and we can revisit once more people have had the chance to try them. In saying that as well should prob at one point look to do ios26Light and ios26Dark as well so we can easily toggle the look per platform but that can be a roadmap item.

**GlassThemeSettings — worth finishing in this PR**

You've done the right thing in applyTo: it now correctly passes frost, rimShade, rimLight, lensModel, frostWeight, and blurWeight through rather than zeroing them when a theme override is applied. The gap is that GlassThemeSettings itself has no optional fields for any of those terms, so there's currently no way to adopt ios27Light globally via GlassTheme — every widget has to set settings: LiquidGlassSettings.ios27Light individually.

If you're up for it in this PR, adding double? frost, double? rimShade, double? rimLight, and GlassLensModel? lensModel to GlassThemeSettings (with matching copyWith and lerp branches, and the ?? propagation through applyTo) would let design systems adopt it with a single theme entry. That's the biggest thing still missing. Happy to discuss the lerp semantics on lensModel being an enum if that's useful — it'd switch at the midpoint as specularSharpness does.

**Capture-path uFrost padding**

One small hygiene item: in liquid_glass_render_object.dart the capture path calls setFloats with only one float for uFrost:

```dart
0.0, // uFrost.x: no frost on the capture path (no cloud rows).
```
uniform vec4 uFrost covers slots 41–44, and while uFrost.x == 0.0 correctly gates out frostAt() today, slots 42–44 are left unset. FragmentShader uniforms persist between calls, so if a uniform is ever inserted before slot 41 in a future refactor, the gate shifts silently and those stale values become load-bearing. Supplying all four floats explicitly costs nothing:

```dart
// No frost on the capture path — all four uFrost components explicit.
0.0, 0.0, 0.0, 1.0,
```
.figma(frost:) — one doc line

Your existing docstring already says frost maps to blur directly — well done for catching that. It might be worth adding one sentence distinguishing it from the new optical frost term, since a developer who reads ios27Light uses frost: 14 and then tries LiquidGlassSettings.figma(frost: 14) will get a very blurry surface with no cloud and no error. Something like:

```dart
/// Note: [frost] here is the Figma blur-slider value — it maps to [LiquidGlassSettings.blur]
/// (a Gaussian sigma), not to [LiquidGlassSettings.frost] (the iOS 27 cloud).
```

Take your time with the theme fields — that's the most involved piece. Everything else is small. Whenever you're ready to take it out of draft we'll get it reviewed and merged quickly. Really looking forward to getting this in. Really appreciate your time on this!!!!

Cheers

JakeThomson:
Thanks, really glad it landed well! Pushed 2c570d4 with all three:

- GlassThemeSettings now has all nine material terms (the four you mentioned plus frostOpacity, frostClamp, frostWeight, blurWeight and rimShadeEnds, since ios27Light needs them all). They go through copyWith, lerp and applyTo, and lensModel switches at the midpoint like specularSharpness.
- The capture path now sets all four uFrost components.
- Added the note to .figma(frost:).

Agreed on keeping it opt-in, and ios26Light/ios26Dark sounds like a good follow-up. I'll take it out of draft once the screenshots are in.

Also have just noticed a slight difference in dark mode styling on lighter backgrounds, will address this tonight in a follow up PR.

